PR #179, opened 2014-04-11
Hungarian localization and tests

gyurisc:
Created the resource file and date, timespan for Hungarian locale. Please, review it and let me know if anything needs to be changed


LBobirca:
Any change we can get a numbers to works for Hungarian?  I have a need for it in writing out check amounts in multiple languages, one of which is Hungary.

gyurisc:
@LBobirca Can you please describe the problem with the numbers to words in Hungarian? I did the original work for the Hungarian localization and I would help you fix this issue if I have a good understanding of what the problem is. Thanks, Krisztian

PR #363, opened 2014-12-19
Use RegexOptions.Compiled if available

hazzik:


MehdiK:
This is now released to NuGet. Thanks for the fix.

Issue #378, opened 2015-02-04
Dehumanize() leaves question mark intact


bornemix:
Maybe this is by design, but:

`"Hey?".Dehumanize()` will equal `"Hey"`
`"Hey ?".Dehumanize()` will equal `"Hey?"`

Tried with `1.33.7`.

MehdiK:
That's a bug. Thanks for reporting.

MehdiK:
Fixed in #381

PR #381, opened 2015-02-10
Fixes trailing question mark reported in #378.

chris-redekop:
A few new tests demonstrate the problem and the fix.

The problem is the upfront single-character check, which lets a single special character through.

The fix is to let the Regex work on every string, and check for empty Enumerations with String.Join and the final ternary operator.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,5 +1,5 @@
 ###In Development
-
+ - [#381](https://github.com/MehdiK/Humanizer/pull/381): Fixes trailing question mark reported in #378.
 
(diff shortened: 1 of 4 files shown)


MehdiK:
Awesome. Thanks for the correction and the unit tests. Could you please add this to the release-notes file?

chris-redekop:
I am glad to help, and thanks for the library!  I added the PR to the release notes.

MehdiK:
Awesome. Thanks for the addition.

MehdiK:
This is now released to NuGet as v1.34.0. Thanks.

Issue #392, opened 2015-03-04
A collection formatter for the current culture has not been implemented yet


Bartmax:
I'm pretty sure this code worked before (data may be different, but still)

var a = new DateTime[] { DateTime.UtcNow, DateTime.UtcNow.AddDays(10) };
var b = a.Humanize(); // THROWS!

A collection formatter for the current culture has not been implemented yet.

I'm not sure what I'm looking into. any help is appreciated.

MehdiK:
Thanks for reporting the issue. It must be related to #386. On reviewing I thought that PR was backward compatible; but apparently I was wrong. What's the *current culture* you're using?

Bartmax:
is the culture set on spanish by chrome. In cultureInfo it prints ES as the culture. hope it helps. Let me know if you need to do something on my end. I'm more than happy to help.

ThomasMentzel:
The problem with the collection formatter is the default implementation, which throws an exception. I had the same issue in German and Danish. This is why I came up with the generalization in my pull request. 
My recommendation is using the "Regular Style" as default collection formatter, which seems the most common usage of concatenating list elements, and overwrite this for those languages which use the oxford style -or even something absolutely different.

I will create a pull request for that and the decision is up to MehdiK.

MehdiK:
Thanks for the PR @ThomasMentzel. I like it. We left you a couple of comments on it.

I'm still wondering why it wasn't failing before and only it started failing after #386!!

MehdiK:
This is now fixed by @ThomasMentzel and available on NuGet as v1.35.0. Thanks for reporting.

hazzik:
Meant to be closed

PR #394, opened 2015-03-10
Default collection formatter not throwing exception

ThomasMentzel:
Regarding issue #392, this implementation of the default collection formatter now does not throw an exception. In my opinion, it is not a good solution to throw an exception in default implementation because this throws an exception on an not supporter language (I had this with German and Danish).


src/Humanizer.Tests/Localisation/DefaultFormatterTests.cs
```diff
@@ -20,5 +21,25 @@ public void TimeSpanHumanizeThrowsExceptionForTimeUnitsLargerThanWeek(TimeUnit t
         {
             Assert.Throws<ArgumentOutOfRangeException>(() => new DefaultFormatter(CultureInfo.InvariantCulture.Name).TimeSpanHumanize(timeUnit, unit));
         }
+
+        [Fact]
+        public void Issue_392_A_collection_formatter_for_the_current_culture_has_not_been_implemented_yet()
+        {
+            var originalCulture = Thread.CurrentThread.CurrentCulture;
```
hazzik: You can use something like:

```csharp
using(new AmbientCulture("es")) {
    // test
}
```

src/Humanizer.Tests/Localisation/DefaultFormatterTests.cs
```diff
@@ -20,5 +21,25 @@ public void TimeSpanHumanizeThrowsExceptionForTimeUnitsLargerThanWeek(TimeUnit t
...
+            var originalUiCulture = Thread.CurrentThread.CurrentUICulture;
+
+            Thread.CurrentThread.CurrentCulture = CultureInfo.CreateSpecificCulture("es");
```
hazzik: I think we need to use different culture, because "es" culture is going to be implemented some day, and the test will break.

src/Humanizer.Tests/Localisation/DefaultFormatterTests.cs
```diff
@@ -20,5 +21,25 @@ public void TimeSpanHumanizeThrowsExceptionForTimeUnitsLargerThanWeek(TimeUnit t
         {
             Assert.Throws<ArgumentOutOfRangeException>(() => new DefaultFormatter(CultureInfo.InvariantCulture.Name).TimeSpanHumanize(timeUnit, unit));
         }
+
+        [Fact]
+        public void Issue_392_A_collection_formatter_for_the_current_culture_has_not_been_implemented_yet()
```
MehdiK: I see where you're going with the method name; but it's inconsistent with the rest of the codebase. Please rename it to something like `HandlesNotImplementedCollectionFormattersGracefully`. You can provide an actual url to the issue on GitHub in a comment.

justin-edwards:
I don't believe this will handle Cyrillic languages,  Asian langues, Arabic, or Greek correctly. There may be others I can't think of at the moment as well.

hazzik:
@justin-edwards what's wrong with Cyrillic languages?

hazzik:
I think `DefaultCollectionFormatter` should return string without any additional separator, just `string.Join(", ", items);`

justin-edwards:
@hazzik, sorry, I had intended to remove Cyrillic languages from that list after I researched it. I may have had bad info on Greek as well (I spoke with someone fluent, but further research contradicts what they said).

hazzik:
@MehdiK, @mexx what'd you say?

mexx:
@hazzik I find the idea of `&` a little better as `,`. We already have other areas where the default going to some English inspired implementation. So I'm OK with `&`.

For me default implementation should produce following output:
```
new[] { "a", "b" }.Humanize() = "a & b"
new[] { "a", "b", "c", "d" }.Humanize() = "a, b, c & d"
```

MehdiK:
I'm with @mexx on this. Also regardless of the default, if one is interested in a better formatter they should just provide their separators in the config.

MehdiK:
I resolved the raised issues, rebased and pushed the code. Thanks.

MehdiK:
This is now released to NuGet as v1.35.0. Thanks for the quick fix.

PR #395, opened 2015-03-19
Changed Humanizer.nuspec to support Xamarin platforms

jkoritzinsky:
I changed the nuspec file to add support for Xamarin platforms.  The project file already supported them because it did not include a profile that excluded them.  With this change, Xamarin users can also install Humanizer from NuGet.


MehdiK:
Good addition. Thanks. Rebased and pushed.

MehdiK:

PR #397, opened 2015-03-26
Add rule for is/are

MatthewSteeples:
I don't know whether this belongs in the same area or not, but we'd like the functionality to extend to the following kinds of sentences:

There is 1 case
There are 5 cases

We can obviously already to "case".ToQuantity(5) to get cases, but we'd like to be able to do the auxiliary verb (is/are) as well.

I've added those entries in but would be happy to get feedback on whether this is the right place or not.

### DIFF
--- a/src/Humanizer.Tests/InflectorTests.cs
+++ b/src/Humanizer.Tests/InflectorTests.cs
@@ -250,6 +250,8 @@ namespace Humanizer.Tests
             yield return new object[] {"wave","waves"};
 
             yield return new object[] {"campus", "campuses"};
+
+            yield return new object[] { "is", "are" };
         }
 
         IEnumerator IEnumerable.GetEnumerator()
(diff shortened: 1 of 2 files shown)


MehdiK:
Thanks @MatthewSteeples. This is alright.

P.S. I was thinking we're not too far from `"There is {quantity} case".ToQuantity(5) => "There are 5 cases"`

MehdiK:
Please add your PR to the releasenotes file and leave me a comment when done and I'll merge. Thanks.

MehdiK:
I added the release notes, rebased and pushed. Thanks.

MehdiK:

MatthewSteeples:
@MehdiK Thanks for merging this. I'll know what to do when we've got other suggestions now. C# 6 does some interesting things with string interpolation that might make your example above a lot more attainable. I've not looked into it much yet.

PR #399, opened 2015-03-27
Support for humanizing DateTimeOffset

### LINKED ISSUE #379: Support for DateTimeOffset
I think a support for DateTimeOffset would be nice complement to the existing DateTime helpers in Humanizer. DateTimeOffset works better in most cases than DateTime, since it represents a very specific moment in time and has better timezone support than the oversimplified DateTimeKind enum in DateTime type.

Please consider.

tompazourek:
I added support for humanizing DateTimeOffset (fixes #379).

It's one of my first pull requests ever, so please comment if I did not do something correctly.

To preserve the backwards compatibility, I did not change the `IDateTimeHumanizeStrategy` interface, but added new `IDateTimeOffsetHumanizeStrategy` interface.

The implementation of these strategies included in the library (`DefaultDateTimeHumanizeStrategy` and `PrecisionDateTimeHumanizeStrategy`) now implement both interfaces, since they share the same algorithm and 99% of the implementation. I thought this was the best solution so as not to introduce compatibility issues.

Support for the `DateTimeOffset` changes the public API (`PublicApiApprovalTest.approve_public_api.approved.txt`) a little bit. I am not sure what is the approval process for the API here, but I updated the API there.

Any feedback and ideas are welcome.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -263,7 +263,7 @@ By default both methods throw a `NoMatchFoundException` when they cannot match t
 In the non-generic method you can also ask the method to return null by setting the second optional parameter to `NoMatch.ReturnsNull`.
 
 ###<a id="humanize-datetime">Humanize DateTime</a>
-You can `Humanize` an instance of `DateTime` and get back a string telling how far back or forward in time that is:
+You can `Humanize` an instance of `DateTime` or `DateTimeOffset` and get back a string telling how far back or forward in time that is:
 
 ```C#
 DateTime.UtcNow.AddHours(-30).Humanize() => "yesterday"
@@ -273,12 +273,13 @@ DateTime.UtcNow.AddHours(30).Humanize() => "tomorrow"
 DateTime.UtcNow.AddHours(2).Humanize() => "2 hours from now"
 ```
 
-Humanizer supports local as well as UTC dates. You could also provide the date you want the input date to be compared against. If null, it will use the current date as comparison base.
+Humanizer supports both local and UTC dates as well as dates with offset (`DateTimeOffset`). You could also provide the date you want the input date to be compared against. If null, it will use the current date as comparison base.
 Also, culture to use can be specified explicitly. If it is not, current thread's current UI culture is used.
 Here is the API signature:
 
 ```C#
 public static string Humanize(this DateTime input, bool utcDate = true, DateTime? dateToCompareAgainst = null, CultureInfo culture = null)
+public static string Humanize(this DateTimeOffset input, DateTimeOffset? dateToCompareAgainst = null, CultureInfo culture = null)
 ```
 
 Many localizations are available for this method. Here is a few examples:
@@ -302,7 +303,10 @@ DateTime.UtcNow.AddMinutes(-40).Humanize() => "40 минут назад"
 There are two strategies for `DateTime.Humanize`: the default one as seen above and a precision based one.
 To use the precision based strategy you need to configure it:
 
-`Configurator.DateTimeHumanizeStrategy = new PrecisionDateTimeHumanizeStrategy(precision = .75)`.
+```C#
+Configurator.DateTimeHumanizeStrategy = new PrecisionDateTimeHumanizeStrategy(precision = .75);
+Configurator.DateTimeOffsetHumanizeStrategy = new PrecisionDateTimeHumanizeStrategy(precision = .75); // configure when humanizing DateTimeOffset
+```
 
 The default precision is set to .75 but you can pass your desired precision too. With precision set to 0.75:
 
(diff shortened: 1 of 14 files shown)


MehdiK:
Hi @tompazourek. Awesome PR and we really needed this.

I am very keen to merge this in; but there are a couple of things I would do differently which I've commented on. Please address these, and leave me a comment and I will merge it.

P.S. I changed the description to say 'Fixes' instead of 'issue'. That will auto close the issue when PR's merged.

readme.md
```diff
@@ -273,12 +273,13 @@ DateTime.UtcNow.AddHours(30).Humanize() => "tomorrow"
 DateTime.UtcNow.AddHours(2).Humanize() => "2 hours from now"
```
MehdiK: Please add an example for `DateTimeOffset` here.

readme.md
```diff
@@ -263,7 +263,7 @@ By default both methods throw a `NoMatchFoundException` when they cannot match t
 In the non-generic method you can also ask the method to return null by setting the second optional parameter to `NoMatch.ReturnsNull`.
 
 ###<a id="humanize-datetime">Humanize DateTime</a>
-You can `Humanize` an instance of `DateTime` and get back a string telling how far back or forward in time that is:
+You can `Humanize` an instance of `DateTime` or `DateTimeOffset` and get back a string telling how far back or forward in time that is:
```
MehdiK: Thanks a lot for adding to the readme.

src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.approve_public_api.approved.txt
```diff
@@ -125,6 +125,7 @@ public class Configurator
 {
     public Humanizer.Configuration.LocaliserRegistry<Humanizer.Localisation.CollectionFormatters.ICollectionFormatter> CollectionFormatters { get; }
     public Humanizer.DateTimeHumanizeStrategy.IDateTimeHumanizeStrategy DateTimeHumanizeStrategy { get; set; }
+    public Humanizer.DateTimeHumanizeStrategy.IDateTimeOffsetHumanizeStrategy DateTimeOffsetHumanizeStrategy { get; set; }
```
MehdiK: FYI the Approver text is a very clean approach to make sure we're not introducing breaking changes. That's all.

src/Humanizer/DateTimeHumanizeStrategy/DefaultDateTimeHumanizeStrategy.cs
```diff
@@ -8,7 +8,7 @@ namespace Humanizer.DateTimeHumanizeStrategy
     /// <summary>
     /// The default 'distance of time' -> words calculator.
     /// </summary>
-    public class DefaultDateTimeHumanizeStrategy : IDateTimeHumanizeStrategy
+    public class DefaultDateTimeHumanizeStrategy : IDateTimeHumanizeStrategy, IDateTimeOffsetHumanizeStrategy
```
MehdiK: I don't think this is a good idea. I think we need to have separate strategy classes for `DateTimeOffset`. I appreciate that the implementation is more or less the same and perhaps you can just delegate the responsibility to the other class or even better extract a class for the algorithm and use it in both implementations.

readme.md
```diff
@@ -302,7 +303,10 @@ DateTime.UtcNow.AddMinutes(-40).Humanize() => "40 минут назад"
 There are two strategies for `DateTime.Humanize`: the default one as seen above and a precision based one.
 To use the precision based strategy you need to configure it:
 
-`Configurator.DateTimeHumanizeStrategy = new PrecisionDateTimeHumanizeStrategy(precision = .75)`.
+```C#
+Configurator.DateTimeHumanizeStrategy = new PrecisionDateTimeHumanizeStrategy(precision = .75);
+Configurator.DateTimeOffsetHumanizeStrategy = new PrecisionDateTimeHumanizeStrategy(precision = .75); // configure when humanizing DateTimeOffset
```
MehdiK: Yeah, I find it weird that you need to configure DateTimeOffset strategy to point to DateTime strategy. We need to create an explicit class for DateTimeOffset.

MehdiK:
This is one clean first PR you've submitted. Awesome work. If you could just please fix the issues I pointed that would be awesome.

Please also add the PR to the releasenotes file. Cheers

tompazourek:
Thanks for the feedback. I separated the implementation of strategies for DateTime and DateTimeOffset, as you suggested. First I experimented with a base class for each pair of strategies, but I felt that the solution was too rigid and the class names were just too long and not nice. So I decided to move the two algorithms to separate `internal static` class. That way the code is very readable and clean, while not polluting the public API.

MehdiK:
Cool. Thanks.

MehdiK:

Issue #388, opened 2015-03-02
Feature Request: Allow MaxUnit and MinUnit for TimeSpan


deadlydog:
First off, this is a great package.  Well done :)

Right now you have the ability to specify a precision integer for TimeSpans.  This is great for hard-coded TimeSpans, but isn't flexible enough for calculated ones.  In my scenario, I want to always have the TimeSpan represented down to the second, whether the time span is for 5 minutes or 5 days.  Ideally I'm looking to be able to do something like:

    (timeSpan1 - timeSpan2).Humanize(minUnit: TimeUnit.Second)

which would generate a string like:
"5 minutes, 15 seconds"
or
"2 days, 3 hours, 2 minutes, 33 seconds"

And to go along with that, being able to specify a maxUnit would be great too, so if I had also specified maxUnit: TimeUnit.Hour, then the above string would have been:
"51 hours, 2 minutes, 33 seconds"

Right now the documentation in the ReadMe.md says that maxUnit exists, but it is not in the latest stable NuGet package, which is currently 1.33.7.  Maybe it's just still in a feature branch and not yet pushed to the NuGet Gallery?

MehdiK:
Thanks for the nice words and suggestion @deadlydog 

Funny you should mention this as [maxUnit was implemented](https://github.com/MehdiK/Humanizer/pull/384/files) just a couple of days ago :) As you said it's not on nuget yet (will be in the next few minutes).

`minUnit` would be a nice addition too. Did you want to send us a PR? It should be relatively easy to implement on top of maxUnit.

alimeziane:
Hi, I would like to work on this subject to participate in your project, that I find very useful. Is anyone treating this evolution or can I start it directly? 
Thanks

MehdiK:
Thanks @blackding. Cool. Go forth and conquer!

danielbrezoi:
Hi, I have worked on this issue and now I'll make a pull request. I have make a version that makes easy to add the implementation of years and months time units if you will want this.

MehdiK:
Closed by #403

PR #403, opened 2015-04-10
Added TimeSpan minUnit option #388

danielbrezoi:
I have add timeSpan minUnit option.


danielbrezoi:
I have refactor the code so that it will be easy to implement the transformation to years and months. This actually is what I want to do next, if it is OK with you, but I don't know how to make the transformation from days. e.g. 1 year = 365 days and 4 years = (3*365) + (1*366) . I need some opinions on this.

MehdiK:

PR #406, opened 2015-04-15
null exception when humanizing enums with [Display Name=""] 

brendan-ssw:
I often use Enums with [Display(Name="xxx")] attributes.
And this usage seems to be encouraged by the docs:
https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.displayattribute.name%28v=vs.110%29.aspx

> The name is typically used as the field label for a UI element that is bound to the property that is annotated with this attribute.

When attempting to Humanize such an Enum, I would get a Null exception from EnumHumanizeExtensions.GetCustomAttribute.

this patch contains a test and fix.

1) detect null from GetDescription call and prevent null pointer exception
2) attempt to lookup GetName if GetDescription fails


src/Humanizer/EnumHumanizeExtensions.cs
```diff
@@ -49,7 +50,17 @@ private static string GetCustomDescription(MemberInfo memberInfo)
...
+                    }
+                    var displayNameMethod = attrType.GetMethod(DisplayAttributeGetNameMethodName);
+                    if (displayNameMethod != null)
+                    {
+                        var obj = displayNameMethod.Invoke(attr, new object[0]);
+                        if (obj != null) return obj.ToString();
+                    }
+                        
                 }
                 var descriptionProperty =
```
mexx: It would be great if you could also prevent the `NullReferenceException` when the value of the custom description property is null.

mexx:
Thanks for the fix, it looks good.
It would be cool if you also take care about the third case of possible `NullReferenceException` source, as mentioned in the code comment.
Please add an entry to release notes.

brendan-ssw:
hi, sorry for the delay. As requested I have:
- added a null check to the third case
- added to release notes

src/Humanizer.sln.DotSettings
```diff
@@ -1,4 +1,6 @@
...
-	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
\ No newline at end of file
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EAddAccessorOwnerDeclarationBracesMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateThisQualifierSettings/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
```
mexx: Hmm, do you know why this file got changed?
Sorry I've didn't recognized this change earlier in the first review.

src/Humanizer/EnumHumanizeExtensions.cs
```diff
@@ -49,14 +50,27 @@ private static string GetCustomDescription(MemberInfo memberInfo)
...
                     attrType.GetProperties()
                         .Where(StringTypedProperty)
                         .FirstOrDefault(Configurator.EnumDescriptionPropertyLocator);
                 if (descriptionProperty != null)
-                    return descriptionProperty.GetValue(attr, null).ToString();
+                {
+                    var obj = descriptionProperty.GetValue(attr, null);
+                    if (obj != null) return obj.ToString();
```
mexx: Could you also add a test for this case?

mexx:
@brendan-ssw No problem about delay, as long as you rebase on HEAD :)
I've left some new comments in the code.

MehdiK:
Any updates on this @brendan-ssw? It's almost there and requires little effort to get it merged.

MehdiK:
Anyone willing to update this PR? Otherwise I might close it as it's open and pending for almost four months now.

MehdiK:
Actually closing now. Please reopen when issues are resolved.

PR #408, opened 2015-04-19
Break out Vocabulary/DefaultVocabulary to make Inflector more extensible.

### LINKED ISSUE #402: Add support for excluding certain user-defined words from pluralization / singularization
One of the places I use Humanizer is on a small recipe site. I pluralize ingredient units where possible (`cup => cups`, `slice => slices`, `lb => lbs`), but some units shouldn't be pluralized. For example, `4 ozs` looks weird, as does `6 tsps`.

Would it be possible to add an overload (or maybe a static property) to the pluralization service so that you can feed it an array of strings that you want to exclude from pluralization?

jkodroff:
fixes #402


jkodroff:
@MehdiK I made this a little less ambitious than what was discussed in #402.  I didn't want to solve anything that wasn't actually being asked for.  The idea here is that:

1. We fixed the immediate problem of the units of measurement.
2. The user can add additional rules if necessary.

jkodroff:
Assuming that the API difference failures are something you'll have to approve.

src/Humanizer/InflectorExtensions.cs
```diff
@@ -22,153 +22,16 @@
...
-using System.Collections.Generic;
 using System.Text.RegularExpressions;
+using Humanizer.Inflections;
 
 namespace Humanizer
 {
     /// <summary>
-    /// Provides hint for Humanizer as to whether a word is singular, plural or with unknown plurality
-    /// </summary>
-    public enum Plurality
```
mexx: @jkodroff what is the reason, why you moved this enum? Please move it back to `Humanizer

src/Humanizer/Inflections/Vocabulary.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        /// <summary>
+        /// Pluralizes the provided input considering irregular words
+        /// </summary>
+        /// <param name="word">Word to be pluralized</param>
+        /// <param name="plurality">Normally you call Pluralize on singular words; but if you're unsure call it with Plurality.CouldBeEither</param>
+        /// <returns></returns>
+        [Obsolete("Use string.Pluralize(bool) instead. This method will be removed in next major release.")]
+        public string Pluralize(string word, Plurality plurality)
```
mexx: Is this method ever used? Please remove if not.

src/Humanizer/Inflections/Vocabulary.cs
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        /// <summary>
+        /// Singularizes the provided input considering irregular words
+        /// </summary>
+        /// <param name="word">Word to be singularized</param>
+        /// <param name="plurality">Normally you call Singularize on plural words; but if you're unsure call it with Plurality.CouldBeEither</param>
+        /// <returns></returns>
+        [Obsolete("Use string.Singularize(bool) instead. This method will be removed in next major release.")]
+        public string Singularize(string word, Plurality plurality)
```
mexx: Is this method ever used? Please remove if not.

src/Humanizer.Tests/InflectorTests.cs
```diff
@@ -21,45 +21,47 @@
...
 
+        // All singular/pluralize tests just need to demonstrate that they're 
+        // delegating to the default vocab.
         [Theory]
-        [ClassData(typeof(PluralTestSource))]
+        [InlineData("mouse", "mice")]
         public void Pluralize(string singular, string plural)
         {
-            Assert.Equal(plural, singular.Pluralize());
+            Assert.Equal(plural, Vocabularies.Default.Pluralize(singular));
```
mexx: Why did you changed the actual test implementation?
Those are inflector extension method tests, they should test them, as they did before.

src/Humanizer.Tests/Inflections/DefaultVocabularyTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            yield return new object[] { "alumnus", "alumni" };
+            yield return new object[] { "fungus", "fungi" };
+
+            yield return new object[] { "wave", "waves" };
+
+            yield return new object[] { "campus", "campuses" };
+
+            yield return new object[] { "is", "are" };
+
+            // Units of measurement:
```
mexx: Please move this addition to the inflector extension method tests.

mexx:
@jkodroff First of all, thank you for the PR.

Regarding the comments:
In Humanizer we try to not break as much as possible and test through the public interface.
You've actually removed the tests for the `Singularize` and `Pluralize` inflector extensions, as stated this should be reverted.
The tests for the `Vocabulary` class should include tests for the new public methods like `AddPlural`, `AddSingular`, `AddIrregular` and `AddUncountable`. Actually this class is tested anyway through the indirect use from inflector extension methods, so I wouldn't necessarily test it.

Regarding the approval:
Please extend the approval file with the added methods as part of your PR.
The build (build.cmd) shouldn't fail neither on your machine nor on the server.

And also add your PR to the release notes.
Also please add documentation for added possibility of vocabulary customization.

Please see [CONTRIBUTING.md](https://github.com/MehdiK/Humanizer/blob/master/CONTRIBUTING.md) for more details.

jkodroff:
@mexx That all sounds reasonable to me.  I think I should be able to find time to make those changes this week.

Thanks for your patience with this new contributor - PR's on well-documented libraries like this one are new to me.

src/Humanizer.Tests/InflectorTests.cs
jkodroff: `Vocabularies.Default` now contains that big list of words, so the big list of words are tested against `Vocabularies.Default`.  `Inflector` now just delegates a call to `Vocabularies.Default`.  (If it were possible to verify that method call cleanly, I'd do that, but testing for equality here is pretty close.)

My original implementation allowed the user to pass a `Vocabulary` to `Pluralize()`, etc, (e.g. a dictionary in another language) but I took this out before submitting the PR as it was overkill for the immediate problem.

I'll be happy to get the tests however they need to be, but I just wanted to make sure my reasoning was clear.  Please let me know how to proceed.

mexx: > ..., but testing for equality here is pretty close.

Not really, the test now verifies only that `Vocabularies.Default` can process **mouse/mice**. It's relation to `Inflector` is gone, as you changed the test. You can change the implementation of the `Inflector` to anything and this test wouldn't break.

dmytro-ielkin:
@jkodroff When do you think you can fix the tests and commit the change?

jkodroff:
Ugh.  I'm so sorry for letting this sit so long.  I'll be sure to get it
done this weekend.

jkodroff:
@mexx Made the requested changes.  We should be good now.  If you want to kill this PR and have me resubmit these changes squashed as a single commit, no problem.

src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.approve_public_api.approved.txt
```diff
@@ -243,6 +243,23 @@ public class In
     public System.DateTime TheYear(int year) { }
 }
 
+public class Vocabularies
+{
+    public Vocabularies() { }
```
mexx: Can you please remove this default constructor from public surface?
This class isn't intended to have any instance, right?

mexx:
@jkodroff Looks better.

> You've actually removed the tests for the `Singularize` and `Pluralize` inflector extensions, as stated this should be reverted.

There were many comments from my side, I think you overlooked this one. Can you please revert the changes to `InflectorTests.cs` file.

jkodroff:
@mexx Are we good now?

src/Humanizer/Inflections/Vocabulary.cs
MehdiK: These `Pluralize` and `Singularize` methods were initially created in Humanizer but then we introduced the other overloads and flagged these as obsolete. There may be users out there who still use these methods. I will put them back.

MehdiK:
Thanks @jkodroff. I made a few changes, rebased your branch on top of origin and pushed up. Please have a look at the changes I made to see if you're happy with them.

MehdiK:

PR #416, opened 2015-05-13
Change BrazilianPortugueseOrdinalizer to PortugueseOrdinalizer

tiagobrito:
Change BrazilianPortugueseOrdinalizer to PortugueseOrdinalizer there is no difference between Brazilian-PortugueseOrdinalizer and Portuguese.

### DIFF
--- a/src/Humanizer/Configuration/OrdinalizerRegistry.cs
+++ b/src/Humanizer/Configuration/OrdinalizerRegistry.cs
@@ -10,7 +10,7 @@ namespace Humanizer.Configuration
             Register("en", new EnglishOrdinalizer());
             Register("es", new SpanishOrdinalizer());
             Register("it", new ItalianOrdinalizer());
-            Register("pt-BR", new BrazilianPortugueseOrdinalizer());
+            Register("pt", new PortugueseOrdinalizer());
             Register("ru", new RussianOrdinalizer());
             Register("tr", new TurkishOrdinalizer());
         }
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,5 +1,6 @@
 ###In Development
-
+ - [#399](https://github.com/MehdiK/Humanizer/pull/416): Change BrazilianPortugueseOrdinalizer to PortugueseOrdinalizer
+ 
 [Commits](https://github.com/MehdiK/Humanizer/compare/v1.35.0...master)
 
 ###v1.35.0 - 2015-03-29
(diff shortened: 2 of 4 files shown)


mexx:
@tiagobrito thanks for the PR. Can you please add it to the `release_notes.md`.

tiagobrito:
@mexx Yes I can, but should I create a tag header: v1.35.1 - 2015-05-12 without existing tag?isn't that weird?
I'm not sure if I get what you meant...

hazzik:
@tiagobrito no, you put it under "In Development" header

tiagobrito:
@hazzik done, thanks for your feedback

src/Humanizer/Configuration/OrdinalizerRegistry.cs
```diff
@@ -10,9 +10,9 @@ public OrdinalizerRegistry() : base(new DefaultOrdinalizer())
             Register("en", new EnglishOrdinalizer());
             Register("es", new SpanishOrdinalizer());
             Register("it", new ItalianOrdinalizer());
-            Register("pt-BR", new BrazilianPortugueseOrdinalizer());
```
MehdiK: Is this not going to break exiting pt-BR users?

tiagobrito: No. pt-PT and pt-BR is the same language, similar to en-US and en-GB. They have some particular differences in some cases, which isn't this case .
When the sub-set of the language is irrelevant we usually use just "pt" or just "en".
To be consistent I've changed the *BrazilianPortugueseOrdinalizer* to *PortugueseOrdinalizer* because the language is Portuguese, there aren't such thing as Brazilian-Portuguese language.

tiagobrito:
Is this forgotten?

hazzik:
@tiagobrito no, it is not forgotten. It LGTM, @mexx @MehdiK.

The one thing, we probably need to do is squash the commits and rebase on top of master. 

Also, not directly related, but just wondering. What if we have "Resources.pt.resx" which contains all Portuguese resources, and "Resources.pt-BR.resx" which contains only Brasil specific wordings. Do you think it will work?

tiagobrito:
@hazzik Yes, that is the best practice, and it will work. but as I told before, right now I didn't find any particular word to justify the existence of a "Resources.pt-BR.resx", but we can have it (empty or duplicated from "Resources.pt.resx") and if one day that case shows up, we can add it there.

MehdiK:
This is now merged. Thanks for the contribution.

MehdiK:
Thanks for the contribution. This is now released to NuGet as v1.37.0.

PR #417, opened 2015-05-21
Support for Serbian language, ToWords()

DeanMilojevic:
I have implemented support for Serbian language, more specific, for now it's only a ToWords() method.


MehdiK:

DeanMilojevic:
No problem, I will continue to implement additional features in my free
time.

PR #429, opened 2015-06-16
Added vNext sample

mrchief:
Added vNext sample and verified that the new provider actually works with vNext.


MehdiK:
Thanks for the contribution @mrchief.

There is one problem with your PR which makes it really hard to review (and also makes me not want to merge it): you have a massive pile of already merged changes in your PR. If you could please pull down the latest from origin to your repository, then make the commits and push a new PR I will review it.

The other thing I would like to add is that although I did initially add the MVC sample in the project I don't think it belongs there. I'd appreciate if you could please exclude your sample. I will need to remove the old one from there too. Happy for you to push that to another repository and then we will link to it from the readme.

Thanks.

mrchief:
Yeah, I merged all commits from master (in hopes that it'll make it easier) but I'll fix this.

Agree about the sample projects. I thought about the same too but in the end just followed your lead. Pushing to another repo is fine or else we can create an Humanizer org and have the different samples as separate repos there (along side the main Humanizer repo). Just a suggestion. I'm fine either which way.

Just glad to be able to help and contribute to this wonderful project!

MehdiK:
Moving the a Humanizer org would be ideal; but it's going to break SO MANY links! I've done that in the past for other projects and made a whole bunch of people cranky!

I'm going to close this PR for now. Looking forward to seeing your fresh clean PR.

Issue #308, opened 2014-06-30
Non portable project


Phyxion:
Can we also get a non portable project? The current portable project doesn't work on Windows XP (and Vista and 7) unless you install the KB2468871 update.

PR #431, opened 2015-06-24
Added a build target for .NET 4.0 Client Profile.

Phyxion:
This is for #308 and is by far the easiest solution without having to copy everything constantly around. If this is accepted, it would be nice to update NuGet as well. I also changed the build.bat to include the 4.0 CP.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,1 +1,2 @@
-@%WINDIR%\Microsoft.Net\Framework\v4.0.30319\msbuild build.proj /m /clp:Verbosity=minimal
+@%WINDIR%\Microsoft.Net\Framework\v4.0.30319\msbuild build.proj /m /clp:Verbosity=minimal /p:TargetPortable="true"
+@%WINDIR%\Microsoft.Net\Framework\v4.0.30319\msbuild build.proj /m /clp:Verbosity=minimal /p:TargetPortable="false"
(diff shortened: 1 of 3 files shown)


MehdiK:
Thanks @Phyxion. I've merged this. The NuGet update would require publishing two separate packages: one for CPL and one for 4.0 CP I believe, right?

Phyxion:
No, you can use one NuGet package with multiple targets.

https://docs.nuget.org/create/creating-and-publishing-a-package#import-msbuild-targets-and-props-files-into-project

MehdiK:
Thanks for the contribution. This is now released to NuGet as v1.37.0. A PR on multiple targets would be nice :)

Phyxion:
I wish I could help but I'm not an expert with NuGet. I do know that several NuGet packages have multiple targets so it should be possible. The NuGet 1.37.0 is still only the PCL so not useful yet for .NET 4.0 clients.

That said, I think it might be possible to just add the 4.0 dll to the "lib\net40" folder as I'm not seeing anything differently in the other files with NuGet packages that have multiple targets. I think this will suffice:
https://github.com/MehdiK/Humanizer/compare/master...Phyxion:patch-1

MehdiK:
It takes a bit more work than that. The build script should move the result of different profiles into different folders which get copied into the NuGet folder structure and then packaged up. This is a needed change and we will tackle it.

Phyxion:
I expected as much, it looked a bit too easy. Looking forward to your commit for this!

Issue #434, opened 2015-06-25
TimeSpan.FromMinutes(50).Humanize(2, null, TimeUnit.Hour, TimeUnit.Minute) is "1 hour, 50 minutes"


GeoffArmstrong:
Obviously this is wrong. `TimeSpan.FromMinutes(50).Humanize(2)` gives the correct `"50 minutes"`.

MehdiK:
Thanks for reporting the problem @GeoffArmstrong. We will try to fix this in the near future. If you have fixed it or would like to do it, please send us a PR and we will merge after review.

GeoffArmstrong:
Thanks! :)

jeremysimmons:
Also having this issue. Humanizer.Localisation.ResourceKeys.TimeSpanHumanize GetResourceKey doesn't support Humanizer.Localisation.Tense the way that DateHumanize does.
Should TimeSpanHumanize support Tense?

MehdiK:
I don't think so @jeremysimmons. `TimeSpan` is not bound to time tense: it's just how long it is without any indication of when; e.g. `TimeSpan.FromMinutes(2)` doesn't say whether it's 2 minutes in the future or in the past: it's just 2 minutes!

jeremysimmons:
@MehdiK the CLR documentation indicates that it does have directionality. https://msdn.microsoft.com/en-us/library/system.timespan(v=vs.110).aspx - A TimeSpan object represents a time interval (duration of time or elapsed time) that is measured as a positive or negative number of days, hours, minutes, seconds, and fractions of a second.

I would say that in most common case, a developer needs the quantity without the sign, but there may be times when they explicitly need the sign.

For now, can the code be patched so that TimeSpans representing a negative quantity will not cause an exception?

I have a changeset I would like to contribute for that fix.

ofirgeller:
Similar problem where a 52 minutes time span becomes 1 hour when max unit is hour.
ts.Humanize(maxUnit: TimeUnit.Hour); // becomes 1 hour.

GeoffArmstrong:
Thanks for the fix!

MehdiK:
The fix for this is now released to NuGet as v1.37.7.

Issue #436, opened 2015-06-29
(DateTime.UtcNow.AddMinutes(-5) - DateTime.UtcNow).Humanize() results in arguments out of range of valid values


htapal:
Just upgraded Humanizer to latest version from 1.34.0 to 1.36.0 and started receiving errors when Humanizing a timespan with a negative value.

> (DateTime.UtcNow.AddDays(-5) - DateTime.UtcNow).Humanize()
Specified argument was out of the range of valid values.
Parameter name: count

htapal:
I downgraded to 1.36.0 and that version doesn't exhibit this issue.  It correctly handles a negative time span as "no time value" message.

MehdiK:
Thanks for reporting this @htapal. Would you be able to send a PR with a supporting test please?

jeremysimmons:
Suggest merging this with #434

MehdiK:
This is fixed by #438 although it will not report the negative values as 'no time value'. It will Humanize it as if it's a positive value.

htapal:
Thanks @MehdiK.  Apologies for the silence as I'd been traveling.  I'm assuming this will be included in the next release?

MehdiK:
No worries. Yup!

MehdiK:
Now up on NuGet

PR #438, opened 2015-07-01
support negative timespans

jeremysimmons:
fixes issue reported in #434


MehdiK:
I don't think this fixes #434. That looks like a logic error. This fixes #436 though. Thanks for the fix.

MehdiK:
Thanks for the contribution. This is now released to NuGet as v1.37.0.

jeremysimmons:
sorry for the wrong tag. thanks for the pull.

PR #440, opened 2015-07-04
fixes #426 adds support for 0 bytes, fixes #434 for TimeSpan < 1hr formatted as 1hr + Xmins

### LINKED ISSUE #426: comma separated bytesize format?
Is there a easy way to convert a number to a comma separated KB string? So that

32768 => "32 KB" and
0 => "0 KB" and
10000000 => "9,765 KB" ?

The only way I can achieve this right now is using
```c#
int i = 32768;
string s = i.Bytes().Kilobytes.ToString("N0") + " KB");
```
Is there a simpler way?

### LINKED ISSUE #434: TimeSpan.FromMinutes(50).Humanize(2, null, TimeUnit.Hour, TimeUnit.Minute) is "1 hour, 50 minutes"

jeremysimmons:
modified the default format string in ByteSize.Format to handle 0 values. Added many unit tests to validate behavior for when Format is explicit with unit (b, B, KB, MB, GB) and implicit using null.

### DIFF
--- a/src/Humanizer.Tests/Bytes/ByteSizeExtensionsTests.cs
+++ b/src/Humanizer.Tests/Bytes/ByteSizeExtensionsTests.cs
@@ -130,6 +130,8 @@ namespace Humanizer.Tests.Bytes
         }
 
         [Theory]
+        [InlineData(0, null, "0 b")]
+        [InlineData(0, "GB", "0 GB")]
         [InlineData(2, null, "2 GB")]
         [InlineData(2, "MB", "2048 MB")]
         [InlineData(2.123, "#.##", "2.12 GB")]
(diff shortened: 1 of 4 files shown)


jeremysimmons:
more changes. would you prefer a pull-request per fix in the future?

MehdiK:
Thanks for the fixes. I prefer smaller specific PRs. That way if one PR gets rejected/abandoned for whatever reason the others could still move through.

MehdiK:
This is now released to NuGet as v1.37.7. Thanks for the contribution.

PR #441, opened 2015-07-05
fixes #246 adding support for placeholder in ByteSize output

jeremysimmons:
breaks backward compatibility by adding placeholder for items with numbers > 1000


MehdiK:
Oh, just saw this. I guess this is not too bad. Still no breaking change though please. We could either have it in configuration or add an optional param to `Humanize` and `ToString` to allow for decimal separators (defaulted to the current behavior to avoid the breaking change).

jeremysimmons:
@Ionian316 closed #426 so this PR is irrelevant.

PR #442, opened 2015-07-08
ToPossessive 

wgimson:
To Possessive functionality added along with tests in new branch ToPossessive, covering cases of default, words ending in 's', common words ending in silent 's' and common possessive pronouns which require no changes. These latter two lists of words will almost certainly be appended to as time goes by.

Just looking at this code I can see that you guys are awesome developers and way beyond my level. Whether you accept this fix or not I welcome any criticism concerning the code and future advice for submitting pull requests and designing bug fixes in general. Thanks for looking!

### DIFF
--- a/src/Humanizer/Humanizer.csproj
+++ b/src/Humanizer/Humanizer.csproj
@@ -151,6 +151,7 @@
     <Compile Include="NoMatchFoundException.cs" />
     <Compile Include="RomanNumeralExtensions.cs" />
     <Compile Include="StringExtensions.cs" />
+    <Compile Include="TransformToPossessive.cs" />
     <Compile Include="ToQuantityExtensions.cs" />
     <Compile Include="Transformer\To.cs" />
     <Compile Include="Transformer\IStringTransformer.cs" />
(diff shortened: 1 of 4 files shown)


src/Humanizer/TransformToPossessive.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            var len = input.Length;
+            var lastChar = input[len - 1];
+            return (char.ToUpper(lastChar) == 'S');
+        }
+
+        private static bool EndsInSilentS(string input)
+        {
+            var uppercaseInput = input.ToUpper().Trim();
+            return (_nounsEndingInSilentS.Contains(uppercaseInput));
```
hazzik: This will fail in turkish.

src/Humanizer/Humanizer.csproj
```diff
@@ -244,9 +245,6 @@
   <ItemGroup>
     <Service Include="{508349B6-6B84-4DF5-91F0-309BEEBAD82D}" />
   </ItemGroup>
-  <ItemGroup Condition=" '$(TargetPortable)' == 'false' ">
```
hazzik: Why this has been removed?

wgimson: Hm good question, it's present on my local system but not in my branched repository. And I don't have any changes to commit.

src/Humanizer/TransformToPossessive.cs
wgimson: Hm reading up on localization now, not really experienced with that.

onovotny:
What is the current status on this? @hazzik? Is this something that's complete and ready? Broken on some languages (Turkish)? Something that's beyond the scope of Humanizer due to it being grammar related?

hazzik:
I think if we accept this, this needs to be moved to a language specific namespace. Something like Humanizer.English. Also, Inflector worth moving to a language specific namespace.

onovotny:
@hazzik Do you think this should go into 2.0 or until the release after that?

hazzik:
@onovotny I think after 2.0

MehdiK:
This could be a great new addition but this PR has been sitting here for over two years @wgimson; so I'm going to close it. It would be great if you could please resolve the conflicts and resubmit the PR. Please also add the new method to readme documentation. cc @onovotny @hazzik

PR #449, opened 2015-08-03
Implement Finnish number to words converter and tests

paavohuhtala:
I've implemented a converter for Finnish language number to words conversion, for both cardinal and ordinal numbers. I've included 50 unit tests, 25 for both kinds.

The implementation is fairly straight-forward, as Finnish numerals are highly systematic. While all of the results produced by the converter are linguistically valid, stylistically perfect conversion is hard if not impossible to do, especially when dealing with large (ordinal) numbers, since they are seldom spelled out.

I modelled the algorithm and the style after the English and the French converters. Cardinal and ordinal numbers are implemented as completely separate methods, which causes some code duplication, but results in cleaner and easier to understand code. There are a few style-related issues I'll have to fix, but other than that I think this feature is pretty much finished.

### DIFF
--- a/src/Humanizer.Tests/Localisation/fi-FI/NumberToWordsTests.cs
+++ b/src/Humanizer.Tests/Localisation/fi-FI/NumberToWordsTests.cs
@@ -1,0 +1,72 @@
+﻿using Xunit;
+using Xunit.Extensions;
+
+namespace Humanizer.Tests.Localisation.fiFI
+{
+    public class NumberToWordsTests : AmbientCulture
+    {
+        public NumberToWordsTests() : base("fi-Fi") { }
+
+        [Theory]
+        [InlineData(0, "nolla")]
+        [InlineData(1, "yksi")]
+        [InlineData(11, "yksitoista")]
+        [InlineData(15, "viisitoista")]
+        [InlineData(19, "yhdeksäntoista")]
+        [InlineData(20, "kaksikymmentä")]
+        [InlineData(25, "kaksikymmentäviisi")]
+        [InlineData(50, "viisikymmentä")]
+        [InlineData(90, "yhdeksänkymmentä")]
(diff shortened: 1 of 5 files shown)


src/Humanizer.Tests/Localisation/fi-FI/NumberToWordsTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        [InlineData(1001, "tuhat yksi")]
+        [InlineData(1234, "tuhat kaksisataakolmekymmentäneljä")]
+        [InlineData(4567, "neljätuhatta viisisataakuusikymmentäseitsemän")]
+        [InlineData(10000, "kymmenentuhatta")]
+        [InlineData(100000, "satatuhatta")]
+        [InlineData(1000000, "miljoona")]
+        [InlineData(10000000, "kymmenenmiljoonaa")]
+        [InlineData(100000000, "satamiljoonaa")]
+        [InlineData(1000000000, "miljardi")]
+        [InlineData(2147483647, "kaksimiljardia sataneljäkymmentäseitsemänmiljoonaa neljäsataakahdeksankymmentäkolmetuhatta kuusisataaneljäkymmentäseitsemän")]  // int.MaxValue
```
MehdiK: Holy, guacamole! ;)

MehdiK:
Thanks for the addition and explanation.

src/Humanizer.Tests/Localisation/fi-FI/NumberToWordsTests.cs
hazzik: @MehdiK check the same number on ordinal :)))

paavohuhtala: @MehdiK I actually got it right the first time, if you can believe that :grinning:

PR #453, opened 2015-08-23
fixes #450

### LINKED ISSUE #450: ByteSize trims number with format string
If you feed 0 to a bytesize method and give it a format string, it will represent them as bits, but not 0 bit but just empty bit.
```csharp
1.Bytes().ToString("#.##") -> "1 B"
0.Bytes().ToString("#.##") -> " b"
0.Bytes().ToString() -> "0 b"
```

Raathigesh:
Fixed "#450 - ByteSize trims number with format string" defect.

### DIFF
--- a/src/Humanizer.Tests/Bytes/ByteSizeExtensionsTests.cs
+++ b/src/Humanizer.Tests/Bytes/ByteSizeExtensionsTests.cs
@@ -332,9 +332,9 @@ namespace Humanizer.Tests.Bytes
 
         [Theory]
         [InlineData(0, null, "0 b")]
+        [InlineData(0, "#.##", "0 b")]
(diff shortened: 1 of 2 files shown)


MehdiK:
Thanks for the fix. A couple of small changes would've made this a better PR:

 - You don't need the issue number in every commit message. Having it in the PR description would be enough.
 - If your description was 'Fixes #450' or 'Fixed #450' merging the PR would close the issue.
 - Squashing your commits into one would be cleaner.

Raathigesh:
I will keep your comments in mind for the next one. This is my first ever PR for an OS project and it was a lot of fun working on this. Thank you for developing and maintaining an awesome project!

PR #459, opened 2015-09-20
Add support of metric numeral expression. fixes #456

### LINKED ISSUE #456: Humanize Count
It seems like there isn't any functionality to humanize a count.  Some examples:

    0 => "0"
    14 => "14"
    71004 => "71k"
    51000492 => "5m" (lower precision)
    51000492 => "5.1m" (higher precision)

It seems very similar to `ByteSize` (maybe `Count`?) except you'd have `Count.Ones`, `Count.Hundreds`, `Count.Thousands`, `Count.Millions`, etc.  `Count.Humanize(precision)` would return the count string at that precision.

Example:

    51000492.ToCount().Humanize(1) => "51000492"
    51000492.ToCount().Humanize(0.999) => "5.1m"

Does this functionality exist in the library?  If not, was it excluded by design?  Or has it just not been thought of or implemented?

A good example of this kind of humanization of a count is on StackOverflow: http://stackoverflow.com/questions?sort=votes  Note how the views ("673k views", "2.4m views", etc.) look.

aloisdg:
Hello,

I created the PR because this is a full working draft. I would be glad to get any kind of reviews.
This is my first time with xUnit (coming from NUnit and MsTest).  Any help is welcome!

Features:

- [x] Convert from metric numeral expression (`double` only)
- [x] Convert to metric numeral expression (`double` only)
- [x] Add `int` support
- [x] Support all symbols (decimal include)
- [x] Support the name of each symbols (see [List of SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes)) (partial => export is supported)
- [x] Support the name of each symbols (see [List of SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes)) (partial => import is supported)
- [x] Handle space between number and symbol
- [ ] Add `ByteSize`'s similar syntax
- [ ] Add the choice of precision
- [ ] Add the support of _small_ symbol / name (this include a switch from char to string in symbols because of deca "da"). Full list : { hecto, h, 1E+2 }, { deca, da, 1E+1 }, { deci, d, 1E-1 }, { centi, c, 1E-2 }. This feature will be lock by default. A boolean in parameter will unlock it.

Also:
- [x] Up-to-date and complete XML documentation
- [x] Include unit tests
- [x] Include unit tests with exception coverage

cheers,

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -30,6 +30,7 @@ Humanizer meets all your .NET needs for manipulating and displaying strings, enu
    - [Number to words](#number-to-words)
    - [Number to ordinal words](#number-to-ordinal-words)
    - [Roman numerals](#roman-numerals)
+   - [Metric numerals](#metric-numerals)
    - [ByteSize](#bytesize)
  - [Mix this into your framework to simplify your life](#mix-this-into-your-framework-to-simplify-your-life)
  - [How to contribute?](#how-to-contribute)
@@ -743,6 +744,27 @@ Also the reverse operation using the `FromRoman` extension.
 "III".FromRoman() => 3
 "IV".FromRoman() => 4
 "V".FromRoman() => 5
+```
+
+###<a id="metric-numerals">Metric numerals</a>
+Humanizer can change numbers to Metric numerals using the `ToMetric` extension. The numbers 1, 1230 and 0.1 can be expressed in Metric numerals as follows:
+
+```C#
+1d.ToMetric() => "1"
+1230d.ToMetric() => "1.23k"
+0.1d.ToMetric() => "100m"
+```
+
+Also the reverse operation using the `FromMetric` extension.
+
+```C#
+1d.ToMetric() => "1"
+1230d.ToMetric() => "1.23k"
+0.1d.ToMetric() => "100m"
+
+"1".FromMetric() => 1
+"1.23k".FromMetric() => 1230
+"100m".FromMetric() => 0.1
 ```
 
 ###<a id="bytesize">ByteSize</a>
--- a/src/Humanizer.Tests/Humanizer.Tests.csproj
+++ b/src/Humanizer.Tests/Humanizer.Tests.csproj
@@ -176,6 +176,7 @@
     <Compile Include="Localisation\zh-Hant\TimeSpanHumanizeTests.cs" />
     <Compile Include="Localisation\zh-CN\DateHumanizeTests.cs" />
     <Compile Include="Localisation\zh-CN\TimeSpanHumanizeTests.cs" />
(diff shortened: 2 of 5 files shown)


aloisdg:
Features:
- [x] Add `int` support
- [x] Support the name of each symbols (see [List of SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes)) (partial. export is supported)
- [ ] Support the name of each symbols (see [List of SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes)) (partial. import is supported)

Also:
- [x] Include unit tests with exception coverage

aloisdg:
For the choice of precision, I am thinking about two ways:

- Full format support:

```c#
public static string ToMetric(this double input, string format,
    bool isSplitedBySpace = false, bool useSymbol = true)
{
    // ...

    return number.ToString(format)
        + (isSplitedBySpace ? " " : String.Empty)
        + GetUnit(symbol, useSymbol);               
}
```

- Precision only support:

```c#
public static string ToMetric(this double input, int n,
    bool isSplitedBySpace = false, bool useSymbol = true)
{
    // ...

    return number.ToString("#." + new string('#', n); // n is the count of numbers
        + (isSplitedBySpace ? " " : String.Empty)
        + GetUnit(symbol, useSymbol);               
}
```

aloisdg:
I think about switching from optional parameters to using overloads methods. I find it easiest to deal with this in public API. Is there any guidance around this in Humanizer? By the way, is there a StyleCop file (or alternative) for the project?

more reading : [1](http://stackoverflow.com/q/251868/1248177) [2](http://stackoverflow.com/q/3316402/1248177)

onovotny:
One favor -- Would you be able to rebase this onto the dev branch? If at all possible, it'd a big help.

How's this looking from a completion perspective otherwise, is it ready for review?

Thanks!

onovotny:
Actually, nm...I'll handle the merge into dev when you're ready. Don't worry about it. Just let me know when you're ready. Thanks!

aloisdg:
Ok thank you for the help. For now, everything works (AFAIk). I think I will let the support of e < 3 open. If someone want it, just create it. For the two others unchecked features, I would love some advices. I think it is a matter of preference.

onovotny:
I've rebased this PR onto https://github.com/MehdiK/Humanizer/tree/aloisdg-metric for review

aloisdg:
Nice. Thanks.

onovotny:
I've merged this in to the dev branch. Thanks for your contribution!

Issue #59, opened 2014-01-14
Getting only needed resources on nuget install


MehdiK:
@hmemcpy raised a valid concern:

> how can I prevent from copying all the language-specific resource to the output dir with Humanizer? Don't want 'em :)

NuGet supports two localization approach: [single & satellite localized packages](http://docs.nuget.org/docs/creating-packages/creating-localized-packages). Humanizer currently uses single localized package & I don't think it's possible to achieve this in this mode.

I think it's a good idea to switch to satellite packages even though it's a breaking change for this and other benefits.

To Do:
  - [x] Create a nupsec for each locale, currently 39
  - [x] Create script to update versions correctly on resource packages
  - [ ] Create script to push all packages to NuGet from the MyGet feed

jeremysimmons:
For what it's worth, I'd like to see the languages separated as separate localization packages too.
Does anyone have a workaround in the meantime?

jeremysimmons:
A workaround for this issue is to add the following to your pre-build event.
```
<PreBuildEvent>forfiles /P &quot;$(SolutionDir)packages\Humanizer.1.36.0\lib\portable-win+net40+sl50+wp8+wpa81+MonoAndroid10+MonoTouch10+Xamarin.iOS10&quot; /c &quot;cmd /c if @isdir==TRUE rmdir /s/q @file&quot; &amp;&amp; EXIT /B 0</PreBuildEvent>
```
The encoded characters are correct. Double Quote and Ampersand have to be XML Escaped inside of the <PreBuildEvent>

onovotny:
Question for everyone asking to split this out -- what's really driving this? I just checked and the localization files are about 325 KB...overall, not huge.

The benefit of the current approach is that "it just works" for all downstream consumers. If you create a library that uses Humanizer, which language dependencies would you choose? Do you force that choice on your consumers who might not know that there are localization packages available?

I guess I'm partially trying to understand what the concern or care is about, what's driving it?

jeremysimmons:
We're a corporate software shop. We cater to US clients that speak ubiquitiusly english. My build engineers freak out whenever a new DLL shows up in the BIN directory for the asp.net app, much less 38 new folders. These folders aren't needed, at all for our scenario which caters exclusively to english, which is already the assembly neutral language. `[assembly: NeutralResourcesLanguage("en")]`

While you're right, the overall size of files is 325 KB, the number of extra files/folders is excessive.
When we want a utility library like this we want one file.

[...]

PR #469, opened 2015-10-29
Refactor packages to enable excluding locales

onovotny:
This PR tracks #59


MehdiK:
Looks good. Thanks @onovotny.

MehdiK:
I am just a bit unsure about `Humanizer.all` name although I'm not sure what else I would call it!

MehdiK:
I am just thinking that when/if there are other packages as part of the library, then `all` implies everything while we just mean all the resources.

onovotny:
So right now, I have `Humanizer` as the meta-package that pull in every Locale and `Humanizer.Core`. Essentially, it preserves current behavior.

For people that don't want all locales, they can just directly install the `Humanizer.Core` package.

Nothing has changed with the dll's or their names, it's all just packaging.

MehdiK:
Nice :+1:

onovotny:
Testing on UWP has turned up an error installing the localized packages...a false error I believe. @Emgarten, can you please assist (email with details should be in your inbox)

onovotny:
@Emgarten: easy repro, check out the `features/integration` branch and try to add `Humanizer.Core.ar` to it via NuGet. You'll see an error about it not being compatible with a UWP app. If you add to the project.json manually, you get that error on build instead.

onovotny:
There's a fix for this but it's dependent on this PR being merged in (and shipped to the public):
https://github.com/NuGet/NuGet.Client/pull/95

Issue #474, opened 2015-11-03
Failing Test: approve_public_api on dev branch, decimal separator issue


rdoubleui:
As suggested in the contributing guidelines I forked the repository and switched to the `dev` branch. I tried to run the tests on the untouched project and hit the following test failure without quite being sure how to resolve it:

```
Test Name:	approve_public_api
Test FullName:	Humanizer.Tests.ApiApprover.PublicApiApprovalTest.approve_public_api
Test Source:	D:\Entwicklung git\Humanizer\src\Humanizer.Tests\ApiApprover\PublicApiApprovalTest.cs : line 16
Test Outcome:	Failed
Test Duration:	0:00:01,506

Result StackTrace:	
bei ApprovalTests.Approvers.FileApprover.Fail()
   bei ApprovalTests.Core.Approver.Verify(IApprovalApprover approver, IApprovalFailureReporter reporter)
   bei ApprovalTests.Approvals.Verify(IApprovalWriter writer, IApprovalNamer namer, IApprovalFailureReporter reporter)
   bei ApprovalTests.Approvals.Verify(IApprovalWriter writer)
   bei ApprovalTests.Approvals.Verify(String text)
   bei Humanizer.Tests.ApiApprover.PublicApiApprovalTest.approve_public_api() in D:\Entwicklung git\Humanizer\src\Humanizer.Tests\ApiApprover\PublicApiApprovalTest.cs:Zeile 30.
Result Message:	ApprovalTests.Core.Exceptions.ApprovalMismatchException : Failed Approval: Received file D:\Entwicklung git\Humanizer\src\Humanizer.Tests\ApiApprover\PublicApiApprovalTest.approve_public_api.received.txt does not match approved file D:\Entwicklung git\Humanizer\src\Humanizer.Tests\ApiApprover\PublicApiApprovalTest.approve_public_api.approved.txt.
```

And I am presented the following diff:

``` 
--- D:/Entwicklung git/Humanizer/src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.approve_public_api.received.txt	Di Nov  3 15:05:10 2015
+++ D:/Entwicklung git/Humanizer/src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.approve_public_api.approved.txt	Di Nov  3 12:33:42 2015
@@ -1028,12 +1028,12 @@ namespace Humanizer.DateTimeHumanizeStrategy
     }
[...]

PR #475, opened 2015-11-04
Added Invariant Culture to rendering of constants

rdoubleui:
Added Invariant Culture to rendering of constants in ApiApprover addressing issue #474


onovotny:
My one concern is that this file is content that's automatically included from the ApiApprover library and that changes will be lost upon upgrade. This fix should also be submitted to it as well.

MehdiK:
What do you think about this @JakeGinnivan?

JakeGinnivan:
Agree, should fix downstream then usage the package. This can be a temp fix but the upgrade will be as smooth

Sent from my Windows Phone
________________________________

rdoubleui:
@onovotny So should this be closed then? Anything I need to do here?

MehdiK:
Thanks for the PR. If you could submit this PR against APIApprover Jake
will merge it in and release it and Humanizer will be fixed through
dependency. Thanks.

rdoubleui:
Ok, did that. Closing it here.

PR #479, opened 2015-11-12
Adding 'never' threshold to DateTimeHumanizeStrategy

### LINKED ISSUE #468: DateTime Humanizations: "never" threshold
I think this is similar to [this](https://github.com/MehdiK/Humanizer/issues/412), a threshold for "never" would be useful for logging purposes, where the DateTime struct sets 01/01/1900 initially. Could this be contributing? I'd like to try to address this.

rdoubleui:
This adds the ability to specify a 'never' threshold, meaning every `DateTime` measured below this value will result in a return value *never* (also added German and French variant to the resources).

Purpose: I use the `DateTime` Humanize extension to display human readable `DateTime` representations on UIs (e.g. "Last Update: 2 minutes ago"). If the `dateNeverThreshold` is sepcified with `DateTime.Now` upon application start, I can circumvent the fact of DateTime being a struct defaulting to *01.01.0001* and resulting in an unnecessary big value (e.g. *2016 years ago*).

Unsure what you think of this, eager to get some feedback.

### DIFF
--- a/src/Humanizer/Localisation/TimeUnit.cs
+++ b/src/Humanizer/Localisation/TimeUnit.cs
@@ -6,6 +6,7 @@
 #pragma warning disable 1591
     public enum TimeUnit
     {
+        Never,
         Millisecond,
         Second,
         Minute,
--- a/src/Humanizer/Localisation/TimeUnit.cs
+++ b/src/Humanizer/Localisation/TimeUnit.cs
@@ -6,7 +6,6 @@
 #pragma warning disable 1591
     public enum TimeUnit
     {
-        Never,
         Millisecond,
         Second,
         Minute,
@@ -14,7 +13,8 @@
         Day,
         Week,
         Month,
-        Year
+        Year,
+        Never
     }
 #pragma warning restore 1591
 }
--- a/readme.md
+++ b/readme.md
@@ -283,10 +283,13 @@ DateTimeOffset.AddHours(1).Humanize() => "an hour from now"
 
 Humanizer supports both local and UTC dates as well as dates with offset (`DateTimeOffset`). You could also provide the date you want the input date to be compared against. If null, it will use the current date as comparison base.
 Also, culture to use can be specified explicitly. If it is not, current thread's current UI culture is used.
+
+In addition when applied to `DateTime` a threshold may be specified (`dateNeverThreshold`). If the measured time is smaller than the threshold, the returned value will default to *never*.
+
 Here is the API signature:
 
 ```C#
-public static string Humanize(this DateTime input, bool utcDate = true, DateTime? dateToCompareAgainst = null, CultureInfo culture = null)
+public static string Humanize(this DateTime input, bool utcDate = true, DateTime? dateToCompareAgainst = null, CultureInfo culture = null, DateTime? dateNeverThreshold = null)
 public static string Humanize(this DateTimeOffset input, DateTimeOffset? dateToCompareAgainst = null, CultureInfo culture = null)
 ```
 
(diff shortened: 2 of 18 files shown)


rdoubleui:
I believe the failed test is related to [this pull request](https://github.com/JakeGinnivan/ApiApprover/pull/13) at [ApiApprover](https://github.com/JakeGinnivan/ApiApprover/). Waiting for the merge as of #474

onovotny:
This looks like an error in the ApiApprover where it's using `XunitReporter` instead of `Xunit2Reporter`. @JakeGinnivan?

rdoubleui:
I didn't accept the changes to `PublicApiApprovalTest.approve_public_api.approved.txt`, I did now, so the build went through ok?!

onovotny:
Looks that way but there's still an underlying NRE in the ApiApprover lib

onovotny:
I started looking over this - did you also make changes to the `DateTimeOffset` versions? I think we should keep the `DateTime` and `DateTimeOffset` versions as close as possible since they should be interchangeable.

rdoubleui:
I have not, understand and I will look into it, if it makes an useful enhancement.

rdoubleui:
Added the same behaviour for `DateTimeOffset`, leading to a merge conflict, though.

src/Humanizer/Localisation/TimeUnit.cs
```diff
@@ -13,7 +13,8 @@ public enum TimeUnit
         Day,
         Week,
         Month,
-        Year
+        Year,
+        Never
```
mexx: I don't know whether extending the `TimeUnit` is a good idea. For me `Never` doesn't fit into this concept.
`TimeUnit` values are used in `TimeSpan` humanization as `minTimeUnit` and `maxTimeUnit`, how would this functionality work when `Never` gets passed into?

mexx:
Currently I don't see the root cause for this feature to exist.
Could you elaborate why the value of `DateTime` struct is `01.01.0001` in first place?
Would it be possible to set this value to the suggested `dateNeverThreshold` or even better to have an optinal `DateTime` value for which we indeed could provide a `Humanize` method, with not having a value being treated as **never**.

src/Humanizer/Localisation/TimeUnit.cs
rdoubleui: I agree, it is a problem passing `Never` into `TimeSpan` humanization, which I didn't notice while working on the `DateTime` humanization.

rdoubleui:
In general the idea was: 

In logging scenarios the humanization of `DateTime` diffs is useful to indicate how long an action is ago, e.g.:
> "Last Update: 2 minutes ago"

If the action has never occured, and the `DateTime` struct has not been initialized, it defaults to `DateTime.MinValue` which leads to a high and non-representative time difference.

While typing this I do realize that you are right: I could achieve the same behaviour by using a `DateTime?` locally:

`LastUpdateString = LastUpdateDateTime.HasValue() ? LastUpdateDateTime.Humanize() : "never";

onovotny:
Does Humanize() on a nullable DateTime/DateTimeOffset value that's currently null return a never string? That might be useful.

rdoubleui:
Thanks, I will look into it and come up with a PR for this. For now I'm closing this PR.

rdoubleui:
@mexx @onovotny Is extending the `IFormatter` interface with a `DateHumanize_Never()` compliant with the idea? And then have it look up the corresponding resource from there?

mexx:
@rdoubleui yes, just like `DateHumanize_Now()`

PR #484, opened 2015-11-19
Fixed tests with decimal separators in test data

sveinfid:
The test ToMetric(input: 1230d) was broken on cultures with comma (,) as decimal separator.
The same with the test FromMetric(input: "1.23k")


onovotny:
Thanks - one thing is that I removed the AmbientCulture class in favor of using an attribute as it ties into xUnit better.

The fix here would be to put `[UseCulture("")]` on the test class

sveinfid:
OK, thanks for the info. I attempted to amend my pull request with a new commit that uses the UseCulture attribute instead. On my machine with comma (,) as decimal separator this works well.

onovotny:
Merged in, thanks!

PR #488, opened 2015-11-25
Updated with vNext sample and readme

### LINKED ISSUE #413: Getting HumanizerMetadataProvider to work in vNext (ASP.NET 5 Sample project)
Seems like `DataAnnotationsModelMetadataProvider` is gone in vNext. Instead there is a `DataAnnotationsMetadataProvider` but it doesn't expose any virtual members to override.

This is my attempt based on the `DefaultModelMetadataProvider` not sure if this is the right way to go:

``` csharp

public class HumanizerMetadataProvider : DefaultModelMetadataProvider
{
    protected override ModelMetadata CreateModelMetadata(DefaultMetadataDetailsCache entry)
    {
        var propertyAttributes = entry.Attributes;
        var modelMetadata = entry.DisplayMetadata;

        if (IsTransformRequired(modelMetadata, propertyAttributes))
            modelMetadata.DisplayName = modelMetadata.SimpleDisplayProperty.Humanize().Transform(To.TitleCase);

        return base.CreateModelMetadata(entry);
    }


    private static bool IsTransformRequired(DisplayMetadata modelMetadata, IReadOnlyList<object> propertyAttributes)
    {
        if (string.IsNullOrEmpty(modelMetadata.SimpleDisplayProperty))
            return false;

        if (propertyAttributes.OfType<DisplayNameAttribute>().Any())
            return false;

        if (propertyAttributes.OfType<DisplayAttribute>().Any())
            return false;

        return true;
    }

    public HumanizerMetadataProvider(ICompositeMetadataDetailsProvider detailsProvider) : base(detailsProvider)
    {
    }
}
```

I still haven't figured out how to plug this in so that it replaces the default.

mrchief:
Resolves #413


onovotny:
Thanks, this is looking a lot better.

A few thoughts: 
  - can the aspnet version be updated to the latest, rc1?
  - there's still a lot of code here as the template added a lot. Can we nuke most of this and have a super-simple page that displays a model that uses humanizer? Like no need for EF, user login, membership, etc.

mrchief:
All good points! I'll send an update.

mrchief:
Cleaned up the examples.

MehdiK:
@onovotny @mrchief is this still relevant? If yes, what's needed before it could be merged?

mrchief:
Been 2 years. I doubt its still accurate. Can't say for sure. Unless ASP.NET core has a new way of doing things, it's still useful.

onovotny:
Yeah, this can be closed. The sample here is based on xproj, which was an old interim format. I think there's value in updated sample, but it should be based on the latest ASP.NET Core templates.

PR #493, opened 2015-12-04
Humanize for DateTime? defaulting to "never"

rdoubleui:
In continuation to #479 I added the `Humanize` overload for `DateTime?`.

PR #500, opened 2015-12-17
Add ToOrdinalWords extension with gender overload

### LINKED ISSUE #328: Ordinal number suffixes
It'd be useful to be able to get just the ordinal suffix, so `st`, `nd`, `rd`, `th` for numbers, especially when working with dates.

Use case:

``` csharp
Assert.That("20th August 2014", Is.EqualTo(new DateTime(2014, 8, 20).<something>());
```

RobPethick:
Fixes #328


src/Humanizer.Tests.Shared/DateToOrdinalWordsTests.cs
```diff
@@ -0,0 +1,25 @@
...
+            Assert.Equal("1st January 2015", date.ToOrdinalWords());
+        }
+
+        [UseCulture("pt-BR")]
+        [Fact]
+        public void CanToOrdinalWordsIncludingGender()
+        {
+            var date = new DateTime(2015, 1, 1);
+
+            Assert.Equal("1º janeiro 2015", date.ToOrdinalWords(GrammaticalGender.Masculine));
```
hazzik: Shouldn't it be automatically?

hazzik:
In portugise, on what the gender of a number is based? Month? Year? Number itself? Or some other word?

src/Humanizer/DateToOrdinalWordsExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+    public static class DateToOrdinalWordsExtensions
+    {
+        /// <summary>
+        /// Turns the provided date into ordinal words
+        /// </summary>
+        /// <param name="input">The date to be made into ordinal words</param>
+        /// <returns>The date in ordinal words</returns>
+        public static string ToOrdinalWords(this DateTime input)
+        {
+            return input.Day.Ordinalize() + input.ToString(" MMMM yyyy");
```
mexx: I wonder whether it works for all locales?
How do Americans write such dates? Wasn't it `January 1st 2015`?

hazzik: Yes, sure, this is language specific.

src/Humanizer.Tests.Shared/DateToOrdinalWordsTests.cs
mexx: @hazzik what do you target with this question, the gender parameter?
I also thought it should be tied to a default value, the `Ordinalize` extensions also do it that way. By this the localization can decide which gender to pick as default.

mexx:
Please also extend the `readme.md` with the documentation of this extension.

src/Humanizer/DateToOrdinalWordsExtensions.cs
RobPethick: Good point, there are plenty of countries that do MMDDYY and plenty that do YYMMDD. I'm working on an update to be localised. Can I ask if I just put up a PR request with only a couple of languages is that ok? Others could then add more in other languages?

mexx: Yes, just provide as minimum the implementation for English, other languages can then localize this functionality with own PRs. See for example #493

src/Humanizer.Tests.Shared/DateToOrdinalWordsTests.cs
hazzik: It's about the [question on the main thread](https://github.com/Humanizr/Humanizer/pull/500#issuecomment-165405118). For example, in russian the gender of the number _in a date_ is always neutral (средний род), and for some languages it can be different. I don't think that we need this method overload at all. 

But, for russian we need an overload with grammatical case, to say "1-го апреля 2016" (_at_ 1st April 2015), for example.

RobPethick:
Should the readme.md use en-GB or en-US as default examples? Or just both?

mexx:
@RobPethick please lean on [number to ordinal words](https://github.com/Humanizr/Humanizer#number-to-ordinal-words) section.

RobPethick:
@mexx thanks but I think in that case en-GB and en-US are the same? And in my case they are different ( 1st January 2015 vs January 1st, 2015)

mexx:
I don't get your point, the documentation should show that this functionality is available. An example in either locale would be sufficient.

RobPethick:
@mexx Thanks! I was just checking there wasn't an existing convention to adhere to.

onovotny:
@RobPethick Just a head's up that it'd be great to resolve this in the next couple of weeks if you'd like to get this into 2.0. I'll leave it to @mexx as to when it's ready.

src/Humanizer/Humanizer.csproj
```diff
@@ -263,4 +264,4 @@
   <Target Name="AfterBuild">
   </Target>
   -->
-</Project>
+</Project>
```
mexx: changes to this file should be reverted, as they only formatting and add no value to the feature.

mexx:
Tagging as vNext as the same thing as @hazzik mentioned for `ToQuantity` applies here:
> I don't think that we can fix this in a couple of weeks. To make it feature complete we need to complete a lot of languages.

RobPethick:
@onovotny @mexx cool I am back at work now after Christmas so should get this finished up soon!

onovotny:
@RobPethick @mexx is this ready to go once the conflicts are resolved? As to support for the other languages, I don't think we need to hold this up -- we can always release again once more languages are added.

onovotny:
rebased and merged here https://github.com/Humanizr/Humanizer/commit/3d71fd90cb7fb45028b9697557ec69358d3366cd, thanks!

PR #504, opened 2015-12-21
Add support for long to ToQuantity

### LINKED ISSUE #503: ToQuantity does not support long quantities
It would be useful if ToQuantity supported long quantities in addition to int. Something like the following would work but will not honor ShowQuantityAs.Words for large positive or negative quantity values:

``` csharp
public static string ToQuantity(this string input, long quantity, ShowQuantityAs showQuantityAs = ShowQuantityAs.Numeric, string format = null, IFormatProvider formatProvider = null)
{
    var str = quantity == 1 ? input.Singularize(false) : input.Pluralize(false);
    if (showQuantityAs == ShowQuantityAs.None)
        return str;
    if ((showQuantityAs == ShowQuantityAs.Numeric) || (quantity > int.MaxValue) || (quantity < int.MinValue))
        return string.Format(formatProvider, "{0} {1}", quantity.ToString(format, formatProvider), str);
    return string.Format("{0} {1}", ((int) quantity).ToWords(), str);
}
```

RobPethick:
Fixes #503

---

This change is [Reviewable](https://reviewable.io/reviews/humanizr/humanizer/504)


mexx:
Sorry, but what's the purpose to extend the domain range by allowing `long` but throw exceptions when this extended range is used? The right fix would be to provide `ToWords` for `long`, also see my [comment](https://github.com/Humanizr/Humanizer/issues/503#issuecomment-166312494)

RobPethick:
@mexx sorry I hadn't seen your comment before putting up my PR. Obviously the point is to allow for it for ShowQuantityAs.None and ShowQuantityAs.Numeric. But I appreciate it kind of seems half complete.

RobPethick:
added in ShowQuantityAs.Words

src/Humanizer.v2.ncrunchsolution
```diff
@@ -0,0 +1,13 @@
+<SolutionConfiguration>
```
mexx: Is this file needed? Isn't it only a configuration for NCrunch?

RobPethick: good catch, I have removed it

mexx:
I don't know should we mention this in the `readme.md` somehow?

RobPethick:
I could, though in general do we discuss types in the readme?

mexx:
You are right, there is no actual notion of the types in the readme.

LGTM, @onovotny can you have a look also and merge if it's all OK?

onovotny:
Rebased and merged here: https://github.com/Humanizr/Humanizer/commit/e88b07ee6a31aa7a0ed1f59e51a1d0edc1b7865a, thanks!

Issue #496, opened 2015-12-11
Add option to convert numbers to different words depending on context


ahjota:
So I was thinking of adding a Korean localization for ToWords() and ToOrdinalWords(), but am having difficulty with the way numbers are handled in Korean in general.

There are two numbering systems, the Korean system and the Chinese system. The Chinese system (1, 2, 3 => 일, 이, 삼) is used for counting money, in dates, and for large quantities. The Korean system (1, 2, 3, => 하나, 둘, 셋) is generally used for counting small quantities. There are other rules that I exclude for brevity, but you get the idea. It's really confusing.

I can't see a way given the current API to register a NumberToWordsConverter that can handle this. Does it make sense to extend the API and overload ToWords() so that it accepts a "context" argument?

var minutesLeft = "15";
minutesLeft.ToWords(Context.Time);

Other ideas would be grand. This is also an issue for other languages, such as Tagalog (which uses Tagalog numbers as well as Spanish-derived numbers).

hazzik:
We can infer the time & date contexts as we have the extensions for these entities. 
The "large vs small" numbers probably can be dealt as well without specifying the context. 
The only thing which we do not have from above is money. The problem is - how many other contexts do exist? What are the "other rules that I exclude for brevity"? 

It might be worth to implement in a separate "Humanizer.Korean" namespace if you can generalize the rules and other people do agree with these rules. Otherwise, it shall be implemented in a client code.

I'm going to close the issue, as there was not much attention given. Feel free to submit the PR.

PR #509, opened 2016-01-14
Add lionbridge translations of fil-PH, ko-KR, ms-MY, and th-TH

onovotny:
I got a free credit for some translation services from Lionbridge and used them to translate the ResX's to four languages not currently supported: 

Filipino
Korean
Malay
Thai

It would be helpful to have a review from someone to validate

### DIFF
--- a/NuSpecs/Humanizer.Core.fil-PH.nuspec
+++ b/NuSpecs/Humanizer.Core.fil-PH.nuspec
@@ -1,0 +1,23 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package>
+  <metadata minClientVersion="2.12">
+    <id>Humanizer.Core.fil-PH</id>
+    <version>$version$</version>
+    <title>Humanizer Locale (fil-PH)</title>
+    <authors>Mehdi Khalili, Oren Novotny</authors>
+    <owners>Mehdi Khalili, onovotny</owners>
+    <projectUrl>https://github.com/Humanizr/Humanizer</projectUrl>
+    <iconUrl>https://raw.github.com/Humanizr/Humanizer/master/logo.png</iconUrl>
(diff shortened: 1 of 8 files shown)


aloisdg:
An idea, would be to export them to csv and ask for a review on reddit (/r/translation for example) or somewhere else on internet.

MehdiK:
This has been here for almost two years. Do you mind if we close it?

hazzik:
@mehdik I think just merge it.

MehdiK:
Can do. There is now a conflict on the branch @onovotny. Could you do a rebase please? Thanks.

onovotny:
@MehdiK just rebased. I didn't merge yet only because there hasn't been any human review of the translations as a sanity check.

onovotny:
wait, the nuspecs are wrong...don't merge yet.

onovotny:
Fixed, should be good now.

MehdiK:
Thanks @onovotny. We have unit tests for all other localisations. This doesn't have unit tests, which is kinda fair, considering it's unverified. Do you think that having tests will help when someone from those languages wants to contribute a fix?

onovotny:
I'm honestly not sure what the right approach is here. Gets down to the purpose of the tests; if it's verifying that the right resources gets returned based on the culture, then perhaps... 

If it's verifying all of the strings, then it may be overkill to have a full suite without any verification.

Maybe a simple test per language that ensures at least one of the correct strings is pulled from the resources?

MehdiK:
Easy to add the tests if and when someone wants to change the provided translations. Thanks for your patience.

justinyoo:
As of the given `Resources.ko-KR.resx`, I confirm that it is fine. However, I'm not sure how #496 has been treated.

Issue #508, opened 2016-01-06
Discuss: Global Config design


onovotny:
Today Humanizer uses a few static configuration options to control formatting/output behavior of some of its methods.

Aside from this design making it harder to run unit tests in parallel, it does place certain restrictions on the library usage - namely that you couldn't have different options depending on the context as they're static.

This may or may not be a real issue but I wanted to start a discussion around it here.

Thoughts?

aloisdg:
I vote for a global config file. I think more flexibility is good thing.

tiesmaster:
Maybe this goes completely against the design of Humanizer, but I'd vote for neither one: don't use static, or configuration, but keep the flexibility with overloads, or constructors when instantiating types.

Why? Although there are other NuGet packages that use static methods, and properties to configure it (like AutoMapper), for me it feels like an old-school approach of providing an API for you as a developer: since they can be accessed from anywhere, they also behave as globals (you never know when, and how things are configured). My experience with that is that it gives a lot of headaches.

These days, dependency injection is becoming the defacto standard, and there are good ways of making things global in that, if you need that when you want to set something global. The same applies to configuration, that doesn't belong in something like a class library like Humanizer.

mexx:
The current design of the API is based upon extension methods, those are static.
For the case of accessibility of the functionality it's OK to have them static. The only thing they should not rely on any mutable shared state, as this brings in the problems mentioned in the issue description.

IMHO we should drop the current support for configuration.
The configuration of operations should be provided as an instance directly to the method performing the operation. For convenience we should provide overload for the common scenarios where the configuration would be populated with some reasonable defaults.

We in generally already take this route, let's take it to the end.

hazzik:
I agree with @mexx.

As I understand actions needed to be performed are:

1. to make Configurator class internal
2. to add overloads where we use Configurator exposed properties.
3. [optional] to add static enum-like strategies for Configurator exposed properties (same as truncator)

onovotny:
Sounds good to me - I'm not a huge fan of static variables so anywhere we can get rid of them is a win to me.

tiesmaster:
:+1:

MehdiK:
The reason I added the static configurator was to allow global configuration without enforcing devs to have to pass the config in every single method. Optional defaults work as long as you want the default value otherwise you will have to pass your override to all method calls. I am also not a big fan of methods with long argument lists which typically happens with a lot of optional params. All that said, I agree with the pain of shared static state.

[...]

PR #511, opened 2016-01-15
Use collection formatter when humanizing timespans. Fixes #497

### LINKED ISSUE #497: Add option to customize list of time indicators of humanized timespans
Probably this title needs some explanation ;) When you humanize a timespan, like this:

````csharp
        const int ONE_WEEK_ONE_DAY = 8;
        var oneWeekOneDayTimeSpan = TimeSpan.FromDays(ONE_WEEK_ONE_DAY);
        var result = oneWeekOneDayTimeSpan.Humanize(2);
````

then you get "1 week, 1 day". However, what I want is to have the option to override the concatenation of the strings, to have an ", and", instead of only a comma between the second to last, and last item, making the output this:

    1 week, and 1 day

[AFAICT](https://github.com/Humanizr/Humanizer/blob/008ce148f814ed0833aee937870c57aed74e16ec/src/Humanizer/TimeSpanHumanizeExtensions.cs#L187), you cannot do this now with Humanizer, and couldn't find an issue for this. I'd suggest adding a `separator` argument to the call, similar to how you can specify that on the [`Humanize` call for collections](https://github.com/Humanizr/Humanizer/blob/008ce148f814ed0833aee937870c57aed74e16ec/src/Humanizer/CollectionHumanizeExtensions.cs#L41). However, I'm not sure if you want another overload for this, besides the [other](https://github.com/Humanizr/Humanizer/blob/008ce148f814ed0833aee937870c57aed74e16ec/src/Humanizer/TimeSpanHumanizeExtensions.cs#L28) [two](https://github.com/Humanizr/Humanizer/blob/008ce148f814ed0833aee937870c57aed74e16ec/src/Humanizer/TimeSpanHumanizeExtensions.cs#L43).

I'm happy to make a PR, if my proposal is accepted, and we can find a good way to add it to the API.

ErikSchierboom:


mexx:
The code looks good to me.
Maybe we could make it non breaking, by defaulting to the old value?

If one would love to have the standard collection format behavior, she can send `null` in.

We can look at the defaults and making breaking when the discussion of global config design in #508 comes to a conclusion and the new way of doing things is clear, what do you think?

ErikSchierboom:
@mexx I have made the requested change, making this PR non-breaking.

mexx:
Sorry, but there is one more thing. Please document this parameter in the readme.md

ErikSchierboom:
@mexx The `readme.md` file has now also been updated.

mexx:
LGTM, @onovotny do you mind to merge it?

PR #517, opened 2016-01-28
allow formatted byte size when humanizing ByteRate

ronnieoverby:
I have needed this on more than one occasion and have had to resort to not using ByteRate because I couldn't format the byte size part of the string.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -937,6 +937,13 @@ text = size.Per(measurementInterval).Humanize(TimeUnit.Hour);
 // 35.15625 GB/hour
 ```
 
+You can specify a format for the bytes part of the humanized output:
+
+```
+19854651984.Bytes().Per(1.Seconds()).Humanize("#.##");
+// 18.49 GB/s
+```
+
 ##<a id="mix-this-into-your-framework-to-simplify-your-life">Mix this into your framework to simplify your life</a>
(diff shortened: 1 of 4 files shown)


mexx:
@ronnieoverby Thank you for your contribution.
Please take a look at the [contribution guideline](https://github.com/Humanizr/Humanizer/blob/dev/CONTRIBUTING.md#contribution-guideline).

The build is failing, could you please fix it.
Please update the `readme.md` with the extended functionality.

ronnieoverby:
I've updated the readme. I'm not sure how I can fix the failing build, though. It seems to be failing when verifying the public api didn't change. I added an overload and didn't change the public api.

Can you help me out?

mexx:
Sure, as you figured out, it fails on public api verification. The public api consist of every publicly available type, method signature and property.
The added overload is part of the public api. You have to update the [approve_public_api.approved.txt](https://github.com/Humanizr/Humanizer/blob/dev/src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.approve_public_api.approved.txt)

ronnieoverby:
Thanks, I didn't realize I had to update that file. For some reason, I was
thinking that the api approver was making sure I didn't remove/change
existing signatures only. I'll give it another shot.

ronnieoverby:
woohoo

onovotny:
Thanks!

PR #522, opened 2016-02-07
Replace dependency (NuGet package ref) with frameworkAssembly

MatthewSteeples:


onovotny:
Thanks for the PR. Unfortunately, as-is, this will not work as it breaks support for .NET Core.

We do not need `frameworkAssembly` here, we really do need the dependencies. Please see my blog post for the detailed explanation:
https://oren.codes/2015/07/29/targeting-net-core/

These dependencies do install on .NET 4.5 and they're harmless. Trying to add empty dependency groups will cause issues down the road as specific TFM > dotnet > portable-*. These packages are the same reference assemblies that live on your computer used by the PCL project system but are now available to dotnet CLI and project.json.

onovotny:
closing as this should be resolved in 2.0.1

PR #530, opened 2016-02-19
Include water and waters as Uncountable for english

matthewvukomanovic:
water and waters are not single and plural of each other.
This includes that as a standard rule for english


matthewvukomanovic:
I'm going to create a enlish-plurals branch for this instead of using dev (that seems to be the requirements instead)

Issue #543, opened 2016-03-17
Add "avant-hier" and "après-demain" to french


aloisdg:
Hello,

There is no support for this two french expressions. It could be nice to implement them.

You can find more informations on [WordReference](http://www.wordreference.com/fren/avant-hier).

I may implement it myself next month. If you want to do it before, go for it! ;)

aloisdg:
In `src/Humanizer.Tests.Shared/Localisation/fr/DateHumanizeTests.cs`:

```C#
[Theory]
[InlineData(1, "hier")]
[InlineData(2, "avant-hier")] // new
[InlineData(10, "il y a 10 jours")]
public void DaysAgo(int days, string expected)
{
        DateHumanize.Verify(expected, days, TimeUnit.Day, Tense.Past);
}

[Theory]
[InlineData(1, "demain")]
[InlineData(2, "après-demain")] // new
[InlineData(10, "dans 10 jours")]
public void DaysFromNow(int days, string expected)
{
        DateHumanize.Verify(expected, days, TimeUnit.Day, Tense.Future);
}
```

To allow this we have to change it in the Resx file `/src/Humanizer/Properties/Resources.fr.resx` :

```xml
<data name="DateHumanize_SingleDayAgo" xml:space="preserve">
    <value>hier</value>
</data>

<data name="DateHumanize_SecondDayAgo" xml:space="preserve">
    <value>avant-hier</value>
</data>

<data name="DateHumanize_SingleDayFromNow" xml:space="preserve">
    <value>demain</value>
</data>

<data name="DateHumanize_SecondDayFromNow" xml:space="preserve">
    <value>après-demain</value>
</data>
```

If we change this, I guess we are will have to change the other resx too. If something like this would fit:

````xml
<data name="DateHumanize_SecondDaysAgo" xml:space="preserve">
    <value>{0} days ago</value>
</data>

<data name="DateHumanize_SecondDaysFromNow" xml:space="preserve">
    <value>{0} days from now</value>
</data>
```

Bascially, SecondDaysFromNow is a copy of `MultipleDays`.

What do you think? @onovotny @hazzik

hazzik:
"Two days from now" and "two days ago". Is ok for me.

aloisdg:
@hazzik Should I implement it for v2.1?

hazzik:
If you can implement this in next day or so, I would say yes. If no - please hold till 2.2/3.0.

aloisdg:
[...]

PR #544, opened 2016-03-17
Add support for avant-hier and après-demain

aloisdg:
See #543 

This PR contains only tests for now. It is a showcase of the feature. Feel free to implement the code to valid them if I don't.


hazzik:
This would be nice to have, but no other language _in_ Humanizer does support this feature.

aloisdg:
@hazzik It should be to hard to implement, shouldn't it? Do you think it could raise some strange issue?

MehdiK:
I'm closing this PR for now as it's quite old and very small. Left a comment on #543 about how we can address this. Please raise a new PR. Thanks.

Issue #548, opened 2016-04-01
french NumberToWords does not take grammatical gender into account


Xiaoy312:
In french, the quantifier should agrees with the noun both in number and **in gender**. However, this rule only applies to number ending with 1's (un) with the exception of 11's (onze) such as 11, 71(60+11) and 91(80+11).

With the expression : 
```
new [] { 1, 21, 31, 41, 51, 61, 81 }
    .Select(x => x.ToWords(GrammaticalGender.Feminine, new CultureInfo("fr-Ca")))
```
What we get | What we expect
--- | ---
un | une 
vingt et un | vingt et une 
trente et un | trente et une 
quarante et un | quarante et une 
cinquante et un | cinquante et une 
soixante et un | soixante et une 
quatre-vingt-un | quatre-vingt-une

Xiaoy312:
Alternatively, this could be used to fix the problem for all dialects of french:

https://gist.github.com/Xiaoy312/0f41c138cdedcddc92a866878eb38240/revisions#diff-05b536df9e00a1303e034dd7e5bbc279R90
```
if (gender == GrammaticalGender.Feminine && lastPart.EndsWith("un"))
    lastPart = Regex.Replace(lastPart, "un$", "une");
```

aloisdg:
Instead of the regex replace, could you just add a 'e' to the end?

```c#
if (gender == GrammaticalGender.Feminine && lastPart.EndsWith("un"))
    lastPart += "e"
```

Xiaoy312:
@aloisdg Indeed, the regex was no longer needed, as previous if-clause already contains the check. It was copy-pasted from a quick fix.

onovotny:
Happy to accept this PR when ready.

PR #552, opened 2016-04-09
fix GrammaticalGender issue #548 with FrenchNumberToWordsConverter

Xiaoy312:
Feminine form of `1, 21, 31, 41, 51, 61, 81` should ends with `une` instead of `un`.


hazzik:
@Xiaoy312 we need tests

hazzik:
@Xiaoy312 is it common French feature, or some regional specific?

aloisdg:
@Xiaoy312 It is a common french feature. We all use it. I think it is mandatory. Source? I am french.

hazzik:
Replaced by #559

Issue #525, opened 2016-02-09
Truncator.FixedNumberOfCharacters [sometimes] counts only digits and letters


hazzik:
1. I think it should be renamed as it does not what it states to do.
2. The logic, when `truncationString` is `null` is different to logic when it is not. For ex.

```csharp
var text = "Text     with more characters than truncate length";
text.Truncate(10, Truncator.FixedNumberOfCharacters); // => "Text     with m…"
text.Truncate(10, string.Empty, Truncator.FixedNumberOfCharacters); // => "Text     with mo"
text.Truncate(10, null, Truncator.FixedNumberOfCharacters); // => "Text     w"
```

mgmccarthy:
@hazzik and @onovotny, I'd like to contribute to Humanizer. This seems like a an issue where I might be able to lend a hand. Just for clarification on this issue, is the problem that you're not expecting that output when the truncationString is null, or that Truncator.FixedNumberOfCharacters should be reanamed? (or both???)

hazzik:
@mgmccarthy I think it is either or both. We need to fix a behaviour and rename, or just fix the behaviour to match the name.

mgmccarthy:
@hazzik, so more specifically, what output do you expect from this call?
`text.Truncate(10, null, Truncator.FixedNumberOfCharacters);`

are you expecting the same output as:
`text.Truncate(10, string.Empty, Truncator.FixedNumberOfCharacters);`

or are you expecting something different?

I want to make sure I understand the problem I'm trying to solve before sending you a PR that is incorrect.

hazzik:
My expectation is to have the same output as `null` in all cases for
FixedNumberOfCharactersTruncator

But also to have FixedNumberOfNonWhitespaceCharacters truncator, which
should match the `string.Empty` current behaviour.

ncksol:
I believe the proper way to fix this is to make the instance with null passed as Truncation String behave the same way when you pass an empty string.
According to description of the class it's suppose to count just characters and digits and whitespace is none of those. It is ignoring it for empty string, but not for null string.
Please see my pr.

onovotny:
@hazzik thoughts? Does the PR address your issue?

onovotny:
Merged in PR

PR #553, opened 2016-04-20
Fixes issue #525

ncksol:
I believe the proper way to fix this is to make the instance with null passed as Truncation String behave the same way when you pass an empty string.
According to description of the class it's suppose to count just characters and digits and whitespace is none of those. It is ignoring it for empty string, but not for null string.


onovotny:
Thanks!

Issue #532, opened 2016-02-19
Issues with German umlauts 


HephaistosGit:
Hello,

there are problems with special german letters:

(Humanizer v.2.0.1, Visual Studio 2013, .NET, 4.6.1)

```
string city = "MÜNCHEN";
Console.WriteLine("{0} => {1}", city, city.Humanize().Transform(To.LowerCase, To.TitleCase));

city = "münchen";
Console.WriteLine("{0} => {1}", city, city.Humanize().Transform(To.LowerCase, To.TitleCase));

city = "München";
Console.WriteLine("{0} => {1}", city, city.Humanize().Transform(To.LowerCase, To.TitleCase));
```

Output should be "München", is:

```
MÜNCHEN => München
münchen => M Nchen
München => Nchen
```

Any suggestions on how I can solve the problem?

HephaistosGit:
Another challenge is the letter "ß" (for this one exists no capital letter)

```
string street = "Mühldorfer Straße 23";
Console.WriteLine("{0} => {1}", street, street.Humanize().Transform(To.LowerCase, To.TitleCase));

street = "MÜHLDORFER strAße 23";
Console.WriteLine("{0} => {1}", street, street.Humanize().Transform(To.LowerCase, To.TitleCase));

street = "mühldorfer STRAßE 23";
Console.WriteLine("{0} => {1}", street, street.Humanize().Transform(To.LowerCase, To.TitleCase));
```

```
Mühldorfer Straße 23 => Hldorfer Stra E 23
MÜHLDORFER strAße 23 => Hldorfer Str E 23
mühldorfer STRAßE 23 => M Hldorfer E 23
```

aloisdg:
@HephaistosGit Could you add this to unit tests with a small PR?

HephaistosGit:
I've added a tiny solution...

```
...
[Test]
public void HumanizeCityTest() {
   const string cityUppercase = "MÜNCHEN";
   const string cityLowercase = "münchen";
   const string cityNormal = "München";
   const string cityExpected = "München";

   // Tests: SUCCESS
   Assert.AreEqual(cityExpected, cityUppercase.ToUpper().Humanize().Transform(To.LowerCase, To.TitleCase));
   Assert.AreEqual(cityExpected, cityLowercase.ToUpper().Humanize().Transform(To.LowerCase, To.TitleCase));
[...]

PR #564, opened 2016-06-15
Modified PascalCaseWordPartsRegex to accept more "word characters"

hangy:
Instead of [A-Z] this uses \p{Lu} (Upper Case Unicode letters) and instead of [a-z] this uess \p{Ll} (lower case Unicode letters). This enables us to use some word characters of different languages, see #532 and 563.

Note that this does not handle the example "mühldorfer STRAßE 23" from #532, but if it were written as "mühldorfer STRAẞE 23", it would be okay.  The previously missing capital ß (ẞ) would require special culture dependent handling, which I do not know if it is wanted in here. However, this deals with "MÜNCHEN" and "ajouté" fine and better than the previous version.

The current `StringHumanizeTests` pass this change, but there might be unexpected changes in production code, as much, much more characters will be accepted and returned in the humanized string.

---
This change is [Reviewable](https://reviewable.io/reviews/humanizr/humanizer/564)

### DIFF
--- a/src/Humanizer.Tests.Shared/StringHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/StringHumanizeTests.cs
@@ -18,7 +18,16 @@ namespace Humanizer.Tests
         [InlineData("?)@", "")]
         [InlineData("?", "")]
         [InlineData("", "")]
+        [InlineData("JeNeParlePasFrançais", "Je ne parle pas français")]
         public void CanHumanizeStringInPascalCase(string input, string expectedResult)
+        {
+            Assert.Equal(expectedResult, input.Humanize());
+        }
+
+        [Theory, UseCulture("tr-TR")]
+        [InlineData("istanbul", "İstanbul")]
+        [InlineData("diyarbakır", "Diyarbakır")]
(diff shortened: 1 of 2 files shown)


jtheisen:
Great! Hope this gets accepted.

hazzik:
The changes LGTM, but can you please add some tests to cover the existing issues, #532 for ex.?

onovotny:
@hazzik is this ready to merge?

hazzik:
Yep

src/Humanizer.Tests.Shared/StringHumanizeTests.cs
```diff
@@ -18,11 +18,20 @@ public class StringHumanizeTests
...
+        [InlineData("JeNeParlePasFrançais", "Je ne parle pas français")]
         public void CanHumanizeStringInPascalCase(string input, string expectedResult)
         {
             Assert.Equal(expectedResult, input.Humanize());
         }
 
+        [Theory, UseCulture("tr-TR")]
+        [InlineData("istanbul", "İstanbul")]
+        [InlineData("diyarbakır", "Diyarbakır")]
+        public void CanHumanizeStringInPascalCaseInTurkish(string input, string expectedResult)
```
hazzik: :+1:

PR #566, opened 2016-06-21
Handling of null, blank and untrimmed collection items

### LINKED ISSUE #555: CollectionHumanizeExtensions.Humanize causes NullReferenceException for null items
This is a headache for me. `.Select(_ => _?.ToString() ?? string.Empty).Humanize()` seems unnecessary; wouldn't it be a sensible default to allow `.Humanize()` directly on collections containing nulls?

Happy to do a PR if you think so.

jnm2:
Fixes #555. I'm guessing [PublicApiApprovalTest.approve_public_api.approved.txt](https://github.com/Humanizr/Humanizer/pull/566/files#diff-c9f04ad944e80f8320d69d564b136694) should be reviewed; are optional params okay or should they be overloads?

### DIFF
--- a/src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
@@ -87,5 +87,17 @@ namespace Humanizer.Tests
             var humanized = _testCollection.Humanize(sc => string.Format("SomeObject #{0} - {1}", sc.SomeInt, sc.SomeString), "or");
             Assert.Equal("SomeObject #1 - One, SomeObject #2 - Two, or SomeObject #3 - Three", humanized);
         }
+
+        [Fact]
+        public void HumanizeHandlesNullItemsWithoutAnException()
+        {
+            Assert.Null(Record.Exception(() => new object[] { null, null }.Humanize()));
+        }
+
+        [Fact]
+        public void HumanizeHandlesNullFormatterReturnsWithoutAnException()
+        {
+            Assert.Null(Record.Exception(() => new[] { "A", "B", "C" }.Humanize(_ => null)));
+        }
     }
 }
--- a/src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using Xunit;
 
@@ -99,5 +100,28 @@ namespace Humanizer.Tests
         {
             Assert.Null(Record.Exception(() => new[] { "A", "B", "C" }.Humanize(_ => null)));
         }
+
+        [Fact]
+        public void HumanizeRunsFormatterOnNulls()
+        {
+            Assert.Equal("1, (null), and 3", new int?[] { 1, null, 3 }.Humanize(_ => _?.ToString() ?? "(null)"));
+        }
+
+        [Fact]
+        public void HumanizeRemovesEmptyItemsByDefault()
+        {
+            Assert.Equal("A and C", new[] { "A", " ", "C" }.Humanize(dummyFormatter));
+        }
+
+        [Fact]
+        public void HumanizeTrimsItemsByDefault()
+        {
+            Assert.Equal("A, B, and C", new[] { "A", "  B  ", "C" }.Humanize(dummyFormatter));
+        }
+
+        /// <summary>
+        /// Use the dummy formatter to ensure tests are testing formatter output rather than input
+        /// </summary>
+        private static readonly Func<string, string> dummyFormatter = input => input;
     }
 }
--- a/readme.md
+++ b/readme.md
@@ -441,7 +441,7 @@ TimeSpan.FromMilliseconds(1299630020).Humanize(3, collectionSeparator: null) => 
 ```
 
 ### <a id="humanize-collections">Humanize Collections</a>
-You can call `Humanize` on any `IEnumerable` to get a nicely formatted string representing the objects in the collection. By default `ToString()` will be called on each item to get its representation but a formatting function may be passed to `Humanize` instead. Additionally, a default separator is provided("and" in English), but a different separator may be passed into `Humanize`.
+You can call `Humanize` on any `IEnumerable` to get a nicely formatted string representing the objects in the collection. By default `ToString()` will be called on each item to get its representation but a formatting function may be passed to `Humanize` instead. Additionally, a default separator is provided ("and" in English), but a different separator may be passed into `Humanize`.
 
 For instance:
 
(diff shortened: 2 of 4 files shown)


src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
```diff
@@ -87,5 +88,58 @@ public void HumanizeUsesObjectFormatterWhenSeparatorIsProvided()
             var humanized = _testCollection.Humanize(sc => string.Format("SomeObject #{0} - {1}", sc.SomeInt, sc.SomeString), "or");
             Assert.Equal("SomeObject #1 - One, SomeObject #2 - Two, or SomeObject #3 - Three", humanized);
         }
+
+        [Fact]
+        public void HumanizeHandlesNullItemsWithoutAnException()
+        {
+            new object[] { null, null }.Humanize();
```
hazzik: Please add check for expected output, and/or that exception is not thrown.

src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
```diff
@@ -87,5 +88,58 @@ public void HumanizeUsesObjectFormatterWhenSeparatorIsProvided()
...
+        }
+
+        [Fact]
+        public void HumanizeHandlesNullFormatterReturnsWithoutAnException()
+        {
+            new[] { "A", "B", "C" }.Humanize(_ => null);
```
hazzik: Same as above

src/Humanizer/CollectionHumanizeExtensions.cs
```diff
@@ -15,46 +15,46 @@ public static class CollectionHumanizeExtensions
         /// using the default separator for the current culture.
         /// </summary>
         /// <returns></returns>
-        public static string Humanize<T>(this IEnumerable<T> collection)
+        public static string Humanize<T>(this IEnumerable<T> collection, StringJoinOptions options = StringJoinOptions.Default)
```
hazzik: For the backward compatibility it is better to add overloads, than defaults.

aloisdg: @hazzik  Can you add more information around this? I think the code will work the same (but I might be wrong). Could it break something? What am I missing?

hazzik: _**TL;DR:** addingition of overload method is non breaking change; addition of a default argument - is._

It's about binary compatibility, if you add overload and do a dll-drop upgrade - everything breaks. You'll think - why would someone do a dll-drop update? To answer this, let's imagine the situation: 

1. Your project depends on Humanizer and other nuget package "A".
2. "A" depends on Humanizer as well.
3. You update the Humanizer, but "A" still references the older version. So, everything breaks.

We are using SemVer. Addition of a new method is a backward compatible change, so we can increase only minor version part. Removal of a method - is not a backward compatible change, so we forced to increase a major version.

hazzik: @aloisdg shorter: if `PublicApiApprovalTest.approve_public_api.approved.txt` has removed lines - it is a breaking change and we are forced to increase a major version part.

hazzik:
>are optional params okay or should they be overloads?
I prefer overloads

src/Humanizer/CollectionHumanizeExtensions.cs
aloisdg: @hazzik Ok thank you!

src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
jnm2: Okay. How do I add a check that an exception is not thrown, other than exactly what I have there?

jnm2:
I was hoping to keep the API simpler and self-documenting, but I understand the binary compat issue would force that to be a major version update.

jnm2:
Also, now `ICollectionFormatter` could have different defaults depending on the implementation which is probably not awesome.

jnm2:
Converted optional params to overloads for minor-version release. Rebased so that you can see, no red in the API.

src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
hazzik: `Assert.DoesNotThrow(()=>...)`

hazzik:
Addition of methods to interface is a breaking change (actually any change to an interface is a breaking change): https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md

jnm2:
Ah. What are our options then?

src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
jnm2: Sure, but I find `DoesNotThrow` redundant and not in a nice way. The whole premise of its existence flies in the face of how the language itself works.

It's on track to be removed from xunit, and I have to say I agree with the sentiments there:  https://github.com/xunit/xunit/issues/188

Other projects like dotnet/roslyn have been removing all usages of `DoesNotThrow`. I'd hate to perpetuate it. Do I have to?

hazzik: Ok. Record.Exception instead of Asser.DoesNotThrow then. I don't care how
it will be implemented, but what I care - is explicitly, that this
particular call to Humanize does not throw

hazzik:
@onovotny @mexx can we just skip 2.1 and release this as 3.0? 

Also, for future, for 3.0 I would like to replace all interfaces we have with default implementations as the classes are more immune to API changes.

src/Humanizer.Tests.Shared/CollectionHumanizeTests.cs
aloisdg: After reading the xUnit.net's issue, I am in favor to `Record.Exception` too. Nice reading by the way :)

jnm2: I don't believe it improves readability- after all, it's a single line in a method named `WithoutAnException` and we all know how the language works- if anything it's subtly misleading, but I'm happy to do whatever matches the project's style.

jnm2:
If 3.0, would you like to see `separator` or `objectFormatter` as optional params?

jnm2:
Hey while we're here, I wonder if anyone has an idea for this? Half the time when I use enumerable.Humanize(), I want to use a verb that needs to be pluralized correctly. (Does/do, is/are, isn't/aren't, exists/exist, etc.)

It's concise to be able to pass a filtered LINQ query to .Humanize(), but not when you have to repeat the query, check if there are fewer or more than one item, and pick the right verb number. Could there be a more idiomatic way of humanizing "{collection} {verb}"?

aloisdg:
@jnm2 It will be a lot of work for english, but good luck for i18n.

onovotny:
Checking in here, is this ready? I'd be great to get 2.1 out soon.

jnm2:
@onovotny It depends. I'd like it in as soon as possible, but things get really ugly if we preserve backwards binary compatibility with `ICollectionFormatter`. If we don't, we're supposed to go to v3.0 which is what @hazzik is asking to do. Thoughts?

onovotny:
I'm not too keen on releasing a `3.0` with just one change like this when there's other things that need to be cleaned up too. Is there really no way to get this without a breaking change? I agree that it'd be good to have default impls to minimize this in the future.

jnm2:
Alternatives:

 1. I don't add `StringJoinOptions` but the behavior changes to trim items and remove blank items and the version stays in 2.x
 2. I do add `StringJoinOptions` to each `ICollectionFormatter` and we go to 3.0- if so, @hazzik would prefer that `ICollectionFormatter` becomes an abstract class to make versioning easier in the future
 3. Version stays 2.x and we add `ICollectionFormatterWithOptions : ICollectionFormatter` and make a mess and also have to figure out if it's safe to change properties that return `ICollectionFormatter` to return `ICollectionFormatterWithOptions`

jnm2:
(I feel the same about 3.0)

 4. We break the rules of binary back-compat

jnm2:
Alternatives:

 1. I don't add `StringJoinOptions` but the behavior changes to trim items and remove blank items and the version stays in 2.x
 2. I do add `StringJoinOptions` to each `ICollectionFormatter` and we go to 3.0- if so, @hazzik would prefer that `ICollectionFormatter` becomes an abstract class to make versioning easier in the future
 3. Version stays 2.x and we add `ICollectionFormatterWithOptions : ICollectionFormatter` and make a mess and also have to figure out if it's safe to change properties that return `ICollectionFormatter` to return `ICollectionFormatterWithOptions`
 4. We break the rules of binary back-compat

I feel the same about 3.0. I'm thinking `StringJoinOptions` could be added in 3.0 but for the time being the default behavior of removing blanks and trimming whitespace would take effect in 2.1. (Option 1) I mean honestly, how often are you going to want to force `A, , C` to happen?

jnm2:
I'm going to rebase again but remove all public API changes so option 1 is ready if that's what you want to go ahead with for 2.1.

jnm2:
Done, rebased on newest from dev.

onovotny:
@hazzik lmk if this PR looks good to merge in for 2.1 and we can add an issue for 3.0 to refactor the interfaces.

Issue #551, opened 2016-04-09
Value cannot be null. Parameter name: format - If language not present.


thedillonb:
Greetings. I'm currently using Humanizer.Core (2.0.1) and found that if a culture is used that is not 'en' (as Core supports) then a 'misleading' exception will be thrown all that way from the `String.Formatter`. Obviously, this is a case of missing resources for the culture, but the error does not make this clear.

If this is not already being addressed I'd be happy to submit a PR.

```
Value cannot be null.
Parameter name: format:   at System.String.Format (System.String format, System.Object[] args) <0x100366bf0 + 0x00058> in <filename unknown>:0 
  at Humanizer.StringExtensions.FormatWith (System.String format, System.Object[] args) <0x100bbb890 + 0x0001b> in <filename unknown>:0 
  at Humanizer.Localisation.Formatters.DefaultFormatter.Format (System.String resourceKey, Int32 number) <0x100bdacd0 + 0x0009b> in <filename unknown>:0 
  at Humanizer.Localisation.Formatters.DefaultFormatter.GetResourceForDate (TimeUnit unit, Tense timeUnitTense, Int32 count) <0x100bdac00 + 0x0005b> in <filename unknown>:0 
  at Humanizer.Localisation.Formatters.DefaultFormatter.DateHumanize (TimeUnit timeUnit, Tense timeUnitTense, Int32 unit) <0x100bdabc0 + 0x0002b> in <filename unknown>:0 
  at Humanizer.DateTimeHumanizeStrategy.DateTimeHumanizeAlgorithms.DefaultHumanize (DateTime input, DateTime comparisonBase, System.Globalization.CultureInfo culture) <0x100bdcfe0 + 0x005a7> in <filename unknown>:0 
  at Humanizer.DateTimeHumanizeStrategy.DefaultDateTimeOffsetHumanizeStrategy.Humanize (DateTimeOffset input, DateTimeOffset comparisonBase, System.Globalization.CultureInfo culture) <0x100bddac0 + 0x0005f> in <filename unknown>:0 
  at Humanizer.DateHumanizeExtensions.Humanize (DateTimeOffset input, Nullable`1 dateToCompareAgainst, System.Globalization.CultureInfo culture) <0x100bbb6b0 + 0x00113> in <filename unknown>:0 
[...]

PR #588, opened 2016-10-02
Add resource value validation for DefaultFormatter

Chrissx:
Added some resource value validation to make the missing resources errors more clear. 

Connected to #551 .

### DIFF
--- a/src/Humanizer/Localisation/Formatters/DefaultFormatter.cs
+++ b/src/Humanizer/Localisation/Formatters/DefaultFormatter.cs
@@ -96,7 +96,7 @@ namespace Humanizer.Localisation.Formatters
             var resourceString = Resources.GetResource(GetResourceKey(resourceKey), _culture);
 
             if (string.IsNullOrEmpty(resourceString))
-                throw new ArgumentException(string.Format("The resource object with key '{0}' was not found"), nameof(resourceKey));
+                throw new ArgumentException(string.Format("The resource object with key '{0}' was not found", resourceKey), nameof(resourceKey));
 
             return resourceString;
         }
@@ -112,7 +112,7 @@ namespace Humanizer.Localisation.Formatters
         {
             var resourceString = Resources.GetResource(GetResourceKey(resourceKey, number), _culture).FormatWith(number);
             if (string.IsNullOrEmpty(resourceString))
-                throw new ArgumentException(string.Format("The resource object with key '{0}' was not found"), nameof(resourceKey));
+                throw new ArgumentException(string.Format("The resource object with key '{0}' was not found", resourceKey), nameof(resourceKey));
 
             return resourceString;
         }
--- a/src/Humanizer/Localisation/Formatters/DefaultFormatter.cs
+++ b/src/Humanizer/Localisation/Formatters/DefaultFormatter.cs
@@ -110,11 +110,12 @@ namespace Humanizer.Localisation.Formatters
         /// <exception cref="ArgumentException">If the resource not exists on the specified culture.</exception>
         protected virtual string Format(string resourceKey, int number)
         {
-            var resourceString = Resources.GetResource(GetResourceKey(resourceKey, number), _culture).FormatWith(number);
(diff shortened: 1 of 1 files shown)


src/Humanizer/Localisation/Formatters/DefaultFormatter.cs
```diff
@@ -88,22 +88,34 @@ private string GetResourceForTimeSpan(TimeUnit unit, int count)
...
+        /// <param name="resourceKey">The resource key.</param>
         /// <returns></returns>
+        /// <exception cref="ArgumentException">If the resource not exists on the specified culture.</exception>
         protected virtual string Format(string resourceKey)
         {
-            return Resources.GetResource(GetResourceKey(resourceKey), _culture);
+            var resourceString = Resources.GetResource(GetResourceKey(resourceKey), _culture);
+
+            if (string.IsNullOrEmpty(resourceString))
+                throw new ArgumentException(string.Format("The resource object with key '{0}' was not found", resourceKey), nameof(resourceKey));
```
aloisdg: You can use string interpolation.

    string.Format("The resource object with key '{0}' was not found", resourceKey)

->

    $"The resource object with key '{resourceKey}' was not found"

src/Humanizer/Localisation/Formatters/DefaultFormatter.cs
```diff
@@ -88,22 +88,34 @@ private string GetResourceForTimeSpan(TimeUnit unit, int count)
...
+        /// <param name="number">The number.</param>
         /// <returns></returns>
+        /// <exception cref="ArgumentException">If the resource not exists on the specified culture.</exception>
         protected virtual string Format(string resourceKey, int number)
         {
-            return Resources.GetResource(GetResourceKey(resourceKey, number), _culture).FormatWith(number);
+            var resourceString = Resources.GetResource(GetResourceKey(resourceKey, number), _culture);
+
+            if (string.IsNullOrEmpty(resourceString))
+                throw new ArgumentException(string.Format("The resource object with key '{0}' was not found", resourceKey), nameof(resourceKey));
```
aloisdg: You can use [string interpolation](https://msdn.microsoft.com/en-us/library/dn961160.aspx).

    string.Format("The resource object with key '{0}' was not found", resourceKey)

->

    $"The resource object with key '{resourceKey}' was not found"

Chrissx: I was not using string interpolation here because if the Stylecop configured to this projects, it can break because you can't configure culture for string interpolation. Fixed now :)

aloisdg: I was not aware of this fact. The more you know 😄

Chrissx: > Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt

I don't think so that was me :D Should I re-trigger?

aloisdg: I checked too and I dont think so, but I cant be sure.

Chrissx: And It's building :)

Chrissx:
:clock1:

aloisdg:
@Chrissx ?

Chrissx:
Ah sorry this was just a ping. Build was succeed. Anything that I need to do?

aloisdg:
You can ping @thedillonb and @onovotny from the original issue.

onovotny:
@aloisdg feel free to merge when you're satisfied

aloisdg:
@onovotny I am not a collaborator 😉

onovotny:
@aloisdg that can be remedied. try again ;)

aloisdg:
@Chrissx Can you add some tests to check the `ArgumentException`? Also, you can fill the return clause (better comments would be nice too, but it is still better than before :) ).

aloisdg:
@onovotny Done! I will try to be worthy. 😄

PR #594, opened 2016-11-01
ByteSize - implement IEEE conformity. fixes #592

### LINKED ISSUE #592: ByteSize does not follow the IEEE 1541 standard
The ByteSize struct uses values such as kilobyte = 1024 byte, whereas the official IEEE 1541 standard dictates that a kilobyte = 1000 bytes, and **kibi**byte = 1024 bytes, and so on. [Wikipedia Link](https://en.wikipedia.org/wiki/Kibibyte).

Also, the SI unit kilo- is normally shortened with a lowercase k, not a capital K.

I will happily fix these issues myself, but it might be seen as a bit of a breaking change, so I thought I'd hear your opinion first.

Alexander882:
Fixed the ByteSize classes so they now conform to the IEEE 1541 standard, i.e. 1 kilobyte = 1000 bytes instead of 1024 bytes. Also added the IEC prefixes; kibi-, mebi-, gibi-, and tebi-, where 1 kibibyte = 1024 bytes etc. in the same manner as the kilo- prefixes worked before.

Added parsing support for negative byte amounts.

Made TryParse return false instead of throwing an exception when given null or whitespace input, so it works like other TryParse-methods.

### DIFF
--- a/src/Humanizer.Tests.Shared/Bytes/ArithmeticTests.cs
+++ b/src/Humanizer.Tests.Shared/Bytes/ArithmeticTests.cs
@@ -38,8 +38,8 @@ namespace Humanizer.Tests.Bytes
         {
             var size = ByteSize.FromKilobytes(2).AddKilobytes(2);
 
-            Assert.Equal(4 * 1024 * 8, size.Bits);
-            Assert.Equal(4 * 1024, size.Bytes);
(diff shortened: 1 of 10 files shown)


MehdiK:
Discussing this on #657

PR #569, opened 2016-06-27
Add Esperanto support to NumberToWords. Fixes #558

aloisdg:
Here is a checklist you should tick through before submitting a pull request: 
- [x] Implementation is clean
- [x] Code adheres to the existing coding standards; e.g. no curlies for one-line blocks, no redundant empty lines between methods or code blocks, spaces rather than tabs, etc.
- [ ] No ReSharper warnings (Is there a global R# file?)
- [x] There is proper unit test coverage
- [x] If the code is copied from StackOverflow (or a blog or OSS) full disclosure is included. That includes required license files and/or file headers explaining where the code came from with proper attribution
- [x] There are very few or no comments (because comments shouldn't be needed if you write clean code)
- [ ] Xml documentation is added/updated for the addition/change (needed?)
- [x] Your PR is (re)based on top of the latest commits from the `dev` branch (more info below)
- [x] Link to the issue(s) you're fixing from your PR description. Use `fixes #<the issue number>`
- [ ] Readme is updated if you change an existing feature or add a new one (Is there a list of language supported?)
- [ ] Run either `build.cmd` or `build.ps1` and ensure there are no test failures (failed but work in VS)

PR #595, opened 2016-11-02
Add new overload for Ordinalize that accepts a CultureInfo

### LINKED ISSUE #593: Add overload for specifying the culture for Ordinalize() extension method
Most extension methods have support for overriding the culture, however `Ordinalize()` doesn't have that. For example, `ToOrdinalWords()` gives you the opportunity to specify it, like this:

````csharp
    10.ToOrdinalWords(new CultureInfo("nl-NL")) => "tiende"
````

It would be nice if you have that overload for `Ordinalize()` as well.

tiesmaster:
This fixes #593, or at least, this is a start with that (@aloisdg asked me to put my preliminary work in a PR).

I've checked all boxes from the checklist, except for the last one (running the build script), since that's not working for me (I get nuget restore errors). I'll drop in the gitter chat to see if I can resolve those.

### DIFF
--- a/src/Humanizer.Tests.Shared/OrdinalizeTests.cs
+++ b/src/Humanizer.Tests.Shared/OrdinalizeTests.cs
@@ -96,14 +96,47 @@ namespace Humanizer.Tests
             var feminineOrdinalized = number.Ordinalize(GrammaticalGender.Feminine);
             Assert.Equal(masculineOrdinalized, feminineOrdinalized);
         }
-        
+
         [Theory]
-        [InlineData("en-US", 1, "1st")]
-        [InlineData("nl-NL", 1, "1e")]
-        public void OrdinalizeWithCultureOverridesCurrentCulture(string cultureName, int number, string ordinalized)
+        [InlineData("en-US", "1", "1st")]
+        [InlineData("nl-NL", "1", "1e")]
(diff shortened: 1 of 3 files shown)


tiesmaster:
@aloisdg Do you have any idea if this is going to be picked up by anyone?

aloisdg:
@tiesmaster I dont know. I am bit out of time to check for the build script. (I had the same problem #569 )

tiesmaster:
@hazzik Thnx for the fix! That's quite embarrassing 😊 It's still failing since the new overload is not in  the `...public_api.approved.txt`. Shall I just add it to it, or is there a different process for that?

hazzik:
@tiesmaster just add.

hangy:
Have you tried rebasing based on the latest dev?

tiesmaster:
@hangy Thnx for the suggestion, I tried that, but didn't work. Finally, I was able to run the API approver locally, and found out that the order is important in the file. So that's fixed now, and the check is 💚  :)

onovotny:
@tiesmaster I believe a few overloads are missing from the public api.txt?

tiesmaster:
@onovotny Ah, forgot the public api.txt again, sorry. Added that, and it passes locally now, so should pass CI.

Had to switch from my Mac to Windows machine, as the public api is not verified on mac 😳.

onovotny:
Thanks!

PR #630, opened 2017-04-27
billions and millions greater than 2

rafaellfontana:
Billions and millions (plural) were only working when the number was greater than 2.
E.g.
1 milhão
2 milhão
3 milhões

 - [x] Implementation is clean
 - [x] No ReSharper warnings


onovotny:
Thanks! Can you please update the unit test to account for this?

src/Humanizer/Localisation/NumberToWords/BrazilianPortugueseNumberToWordsConverter.cs
```diff
@@ -26,7 +26,7 @@ public override string Convert(int number, GrammaticalGender gender)
             if ((number / 1000000000) > 0)
             {
                 // gender is not applied for billions
-                parts.Add(number / 1000000000 > 2
+                parts.Add(number / 1000000000 == 1
                     ? string.Format("{0} bilhões", Convert(number / 1000000000, GrammaticalGender.Masculine))
                     : string.Format("{0} bilhão", Convert(number / 1000000000, GrammaticalGender.Masculine)));
```
aloisdg: You can use interpolated string:
```c#
var genderMark = number / 1000000000 == 1 ? "ões" : "ão";
parts.Add($"{Convert(number / 1000000000, GrammaticalGender.Masculine)} bilh{genderMark}");
```

And you can use something similar for millions (or even make it a private static method).

PR #632, opened 2017-05-03
Fuzzy month years logic for time span humanize extensions

MaStr11:
Fixes #583
This change now allows calls to `TimeSpanHumanizeExtensions.Humanize` with the parameter `maxUnit` and `minUnit` set to `TimeUnit.Month` or `maxUnit: TimeUnit.Year`. The duration in month and years is approximated based on 365.2425 days a year (length of the [gregorian calendar](https://en.wikipedia.org/wiki/Gregorian_calendar)). Therefore the month are alternating between 30 and 31 days and the years are alternating between 365 and 366 days.

The existing default value for `maxUnit=TimeUnit.Week` is kept because weeks are precise while month and year are not.

Most of the changes are related to the translation of the month and year identifiers. All text were translated with google translator except:

* *German* I'm a native German.
* *Russian* Reviewed by a native speaker.
* *fi-FI* Tests and translations for timespan were already missing.
* *el* Tests and translations for timespan were already missing.

I marked the test related to translations with the `[Trait("Translation", "Google")]` attribute to make clear which translations are of poor quality.

Checklist for the Translation Reviews:

- [ ] af
- [ ] ar
- [ ] bg
- [ ] bn-BD
- [ ] cs
- [x] da @khellang
- [x] de @MaStr11
- [ ] el
- [ ] es
- [x] fa @ctyar 
- [ ] fi-FI
- [ ] fr-BE
- [ ] fr
- [x] he @M-Zuber 
- [ ] hr
- [ ] hu
- [ ] id
- [ ] it
- [ ] ja
- [x] nb-NO @khellang 
- [x] nb @khellang 
- [x] nl @NinoFloris
- [ ] pl
- [ ] pt
- [ ] ro
- [x] ru @herecydev 
- [ ] sk
- [ ] sl
- [ ] sr-Latn
- [ ] sr
- [x] sv @khellang 
- [ ] tr
- [ ] uk
- [ ] uz-Cyrl-UZ
- [ ] uz-Latn-UZ
- [ ] vi
- [x] zh-CN @jeremymeng 
- [x] zh-Hans @jeremymeng 
- [x] zh-Hant @jeremymeng

PR #636, opened 2017-05-05
Added tests for Greek deleted obsolet entries in resx. Corrected Roma…

MaStr11:
…nian and marked both with "Native speaker"
Corrections of the Romanian language as reviewed by @ds-b in #632


onovotny:
There's a lot of deletes in there -- are we not using those anymore where did they go and how about the other languages, do they also have those?

MaStr11:
The deleted entries are not used by by the [DefaultFormatter](https://github.com/Humanizr/Humanizer/blob/dev/src/Humanizer/Localisation/Formatters/DefaultFormatter.cs) used by the Greek language. Other languages use special rules (e.g. [RussianFormatter](https://github.com/Humanizr/Humanizer/blob/dev/src/Humanizer/Localisation/Formatters/RussianFormatter.cs) appends "_Paucal" (for number 2 to 4) or _"Singular" (for numbers ending with 1)). Most languages like [German](https://github.com/Humanizr/Humanizer/blob/dev/src/Humanizer/Properties/Resources.de.resx) don't have the entries ending with _Above20, _Dual, _Plural, _Singular, _TrialQuadral. All entries I deleted were duplicates, so actually no information is lost.

onovotny:
Thanks for the explanation. We ought to have a Wiki entry on what's needed for different languages.

PR #638, opened 2017-05-08
Use CultureInfo.Parent to find fallback CultureInfo.

### LINKED ISSUE #626: zh-Hant and zh-Hans do not resolve to zh-CN and zh-HK
We use the mainland China and Hong Kong cultures to indicate the languages. Unfortunately Humanizer only supports zh-Hant and zh-Hans and treats zh-HK and zh-CN as if they are English. They should be mapped to the traditional and simplified cultures.

hazzik:
fixes #626

### DIFF
--- a/src/Humanizer.Tests.Shared/Humanizer.Tests.Shared.projitems
+++ b/src/Humanizer.Tests.Shared/Humanizer.Tests.Shared.projitems
@@ -153,6 +153,8 @@
     <Compile Include="$(MSBuildThisFileDirectory)Localisation\zh-CN\TimeSpanHumanizeTests.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Localisation\zh-Hans\DateHumanizeTests.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Localisation\zh-Hans\TimeSpanHumanizeTests.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Localisation\zh-HK\DateHumanizeTests.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Localisation\zh-HK\TimeSpanHumanizeTests.cs" />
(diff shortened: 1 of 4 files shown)


MaStr11:
@hazzik That PR looks really good. I only was wondering why you have chosen to have zh-Hans and zh-Hant as supported locals? I would have expected to have *zh-Hant* and *zh*. *zh* would be the fall-back if anything else is failing. That seems more robust to me. On the other hand:
*A neutral culture is a culture that is associated with a language but not with a country/region. A specific culture is a culture that is associated with a language and a country/region. For example, fr is the neutral name for the French culture, and fr-FR is the name of the specific French (France) culture. **Note that Chinese (Simplified) and Chinese (Traditional) are also considered neutral cultures.***
[Source](https://msdn.microsoft.com/de-de/library/system.globalization.cultureinfo(v=vs.110).aspx#Invariant)
Humanizer is very inconsistent in regards to specifying neutral cultures:
* bn-BD: no neutral language given
* fi-FI: no neutral language given
* fr, fr-BE: neutral language given
* nb, nb-NO: neutral language given
* sr, sr-Lat: neutral language given
* uz-Cyrl-UZ, uz-Lat-UZ: no neutral language given
* zh: no neutral language given
Should we create an issue to address this?

hazzik:
>I only was wondering why you have chosen to have zh-Hans and zh-Hant as supported locals?

It seems that there are no locales which bypass these two neutral cultures directly to "zh". So, unless anyone specifically set culture as "zh" we don't have any issues.

Not sure if I could answer other questions:)

hazzik:
I'll amend the PR to exclude zh-CN depreciation. We can groom cultures later.

hazzik:
Removed depreciation of zh-CN culture

hazzik:
I'll wait the build to finish, then merge.

PR #657, opened 2017-09-27
Byte size to follow IEEE standard

Alexander882:
Changed ByteSize to follow IEEE and IEC standards for byte sizes (kibi vs kilo, etc.) see #592

### DIFF
--- a/src/Humanizer.Tests.Shared/Bytes/ArithmeticTests.cs
+++ b/src/Humanizer.Tests.Shared/Bytes/ArithmeticTests.cs
@@ -38,8 +38,8 @@ namespace Humanizer.Tests.Bytes
         {
             var size = ByteSize.FromKilobytes(2).AddKilobytes(2);
 
-            Assert.Equal(4 * 1024 * 8, size.Bits);
-            Assert.Equal(4 * 1024, size.Bytes);
+            Assert.Equal(4 * 1000 * 8, size.Bits);
+            Assert.Equal(4 * 1000, size.Bytes);
             Assert.Equal(4, size.Kilobytes);
         }
 
@@ -48,9 +48,9 @@ namespace Humanizer.Tests.Bytes
         {
             var size = ByteSize.FromMegabytes(2).AddMegabytes(2);
 
-            Assert.Equal(4 * 1024 * 1024 * 8, size.Bits);
-            Assert.Equal(4 * 1024 * 1024, size.Bytes);
-            Assert.Equal(4 * 1024, size.Kilobytes);
+            Assert.Equal(4 * 1000 * 1000 * 8, size.Bits);
+            Assert.Equal(4 * 1000 * 1000, size.Bytes);
+            Assert.Equal(4 * 1000, size.Kilobytes);
             Assert.Equal(4, size.Megabytes);
         }
 
@@ -59,10 +59,10 @@ namespace Humanizer.Tests.Bytes
         {
             var size = ByteSize.FromGigabytes(2).AddGigabytes(2);
 
-            Assert.Equal(4d * 1024 * 1024 * 1024 * 8, size.Bits);
-            Assert.Equal(4d * 1024 * 1024 * 1024, size.Bytes);
-            Assert.Equal(4d * 1024 * 1024, size.Kilobytes);
-            Assert.Equal(4d * 1024, size.Megabytes);
+            Assert.Equal(4d * 1000 * 1000 * 1000 * 8, size.Bits);
+            Assert.Equal(4d * 1000 * 1000 * 1000, size.Bytes);
+            Assert.Equal(4d * 1000 * 1000, size.Kilobytes);
+            Assert.Equal(4d * 1000, size.Megabytes);
             Assert.Equal(4d, size.Gigabytes);
         }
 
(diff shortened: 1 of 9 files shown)


MehdiK:
Thanks for the great contribution. This is a significant breaking change though, and like @omar mentioned [here](https://github.com/Humanizr/Humanizer/issues/592#issuecomment-262893703), while technically wrong, is what most people expect.

Can I suggest that you keep the existing functionality as default behavior, to avoid breaking the behavior, and add a strategy to switch to IEEE standard through config? [Here]
(https://github.com/Humanizr/Humanizer/blob/dev/src/Humanizer/Configuration/Configurator.cs#L113) is an example of a similar implementation.

glen-84:
@clairernovotny If the conflicts were resolved, would this PR be accepted?

glen-84:
@SimonCropp

- The IEC Standard has been adopted by the IEEE, EU, ISO, and NIST, since 2009.
- `Kilo-` literally means "thousand", `Mega-` is used in [many places](https://en.wikipedia.org/wiki/Mega-#Common_examples_of_usage) to refer to 1 million, etc.
- If it's 1,024 bytes by default, then both kilo- and kibi- would be 1,024 bytes.
- The [ByteSize](https://github.com/omar/ByteSize) library (authored by Omar, referenced above) [switched](https://github.com/omar/ByteSize?tab=readme-ov-file#ratio-changes-huge-breaking-change) to SI units.
- It should mean one thing. When you ask someone how many bytes in a megabyte, they shouldn't answer "it depends".

IMO, the default should be decimal units (1 kB = 1,000 bytes).

SimonCropp:
@MehdiK  so @glen-84 makes some compelling points. especially since the [ByteSize](https://github.com/omar/ByteSize) lib has switched over. thoughts?

MehdiK:
That is a fair point and I completely agree but there are folks who have taken dependency on this - changing the behaviour will break many apps. If we want to change the behaviour, we need to bump the major version. I still suggest we have a strategy in there that allows people to revert back to the old behaviour through config. Thoughts?

glen-84:
> If we want to change the behaviour, we need to bump the major version.

Absolutely.

> I still suggest we have a strategy in there that allows people to revert back to the old behaviour through config. Thoughts?

I have no issue with that. I just believe that the _default_ should be decimal units.

SimonCropp:
> I still suggest we have a strategy in there that allows people to revert back to the old behaviour through config. Thoughts?

@MehdiK i think it would suffice to tell people to move to the new extensions methods. eg "if you want the old behavior change usages of Kilobytes to Kibibytes"

glen-84:
> if you want the old behavior change usages of Kilobytes to Kibibytes

They may still want it to show "kilobytes" and not "kibibytes" in certain cases:

> The binary interpretation of metric prefixes is still prominently used by the [Microsoft Windows](https://en.wikipedia.org/wiki/Microsoft_Windows) operating system.[[9]](https://en.wikipedia.org/wiki/Kilobyte#cite_note-auto-9) Binary interpretation is also used for [random-access memory](https://en.wikipedia.org/wiki/Random-access_memory) capacity, such as main memory and [CPU cache](https://en.wikipedia.org/wiki/CPU_cache) size, due to the prevalent [binary addressing](https://en.wikipedia.org/wiki/Binary_address) of memory.

SimonCropp:
@glen-84 

> They may still want it to show "kilobytes" and not "kibibytes" in certain cases:

and they can. there are two extension methods for each unit. take a look at this pr https://github.com/Humanizr/Humanizer/pull/1413

CollinAlpert:
> I think 1,024 bytes will still return KB.

@glen-84 That is correct and I think this should be the default anyway. If you want a KiB output, you can specify so in the format by passing "KiB" to the call.

glen-84:
But what does this output?

```cs
var maxFileSize = (10).Kibibytes();

maxFileSize.LargestWholeNumberSymbol; // ?
```

I'd expect to see `KiB`, but I think it'll be `KB`?

CollinAlpert:
@glen-84 I agree this would be the expectation at first glance. However a more consistent approach would be not to think of kilobytes and kibibytes as separate units, but more of different representations of the actual unit (byte). The same amount of bytes can be expressed in KB as well as KiB. These are not different units.

In your example, I personally would expect "KB" to be printed, since it is more common and many people don't even know what a kibibyte is. For those people who want to have "KiB" displayed, they can still request it. 

Also, the implementation would become quite bloated when trying to guess what the developer wants. Say, you write the following: `(10).Kibibytes() + (10).Kilobytes()`? Does it still make sense to have `LargestWholeNumberSymbol` print "KiB"?

clairernovotny:
Thank you for the substantial work and discussion on standards-compliant byte sizes. The newer PR #1413 became the follow-up implementation for this direction, so I am closing this older duplicate while retaining the design history here.

clairernovotny:
Thank you again, @Alexander882, for starting this work and discussion. The compatibility-safe explicit IEC support has now landed on `main` in #1851 for the next release, carrying forward and crediting @CollinAlpert's later implementation in #1413.

Existing KB-TB behavior remains unchanged; #592 stays open for any future release-policy decision about redefining those APIs as decimal SI units.

Issue #663, opened 2017-10-23
Camelize() fails on empty/null strings


jpobst:
AOORE with this code:

`"".Camelize()`

ANE with this code:

`((string)null).Camelize()`

hangy:
I think that most of Humanizer's string extensions will throw an ArgumentNullException for null strings, so that could be expected for `Camelize`, too. The ArgumentOutOfRangeException needs to be prevented, though.

MehdiK:
Fixed in #664. Thanks.

PR #664, opened 2017-10-24
Fix ArgumentOutOfRangeException for Camelize

hangy:
This ensures that Camelize does not throw an unexpected ArgumentOutOfRangeException if the input string is empty. See #663.


MehdiK:
Thanks for the fix.

PR #675, opened 2017-11-30
Annotate source code with JetBrains.Annotations

Nihlus:
This PR implements wide-coverage source code annotations using the JetBrains.Annotations package. In doing so, static inspection accuracy of source code using Humanizer with ReSharper and Rider is vastly improved.

In order to maintain a small footprint and not include redundant dependencies, JetBrainsAnnotations.Fody is used to take the annotated assembly and generate external annotations at compile time. This removes the dependency on JetBrains.Annotations, and allows Humanizer to ship an external annotation file instead.

No code changes have been made. Nullability annotations have been written very conservatively, and only in cases where null values are explicitly handled in code have they been annotated as allowing such. These instances have also been annotated with contract annotations.

### DIFF
--- a/src/Humanizer/Bytes/ByteRate.cs
+++ b/src/Humanizer/Bytes/ByteRate.cs
@@ -38,6 +38,7 @@ namespace Humanizer.Bytes
         /// </summary>
         /// <param name="timeUnit">Unit of time to calculate rate for (defaults is per second)</param>
         /// <returns></returns>
+        [Pure]
         [NotNull]
         [PublicAPI]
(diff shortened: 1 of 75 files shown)


Nihlus:
I've rebased the PR on the master branch, and added some more annotations to new or changed code.

Nihlus:
I'm unsure why the unit tests are failing on VSTS - locally, everything is running fine.

onovotny:
Closing due to age. Going forward, we'll use C# 8 non-nullables for this as well. That should work everywhere.

PR #681, opened 2017-12-30
[WIP] fixes #651

### LINKED ISSUE #651: Timespan.humanize() shows weird results
Hi, 

An application I had to take over uses Humanizer and is showing strange results in some cases when using a Timespan.
I'm not familiar with Humanizer and I tried finding a similar issue and/or solution online but there was none so here I am.

'when using this code and with a timespan of for example 45 days the humanized text results in descriptions like  **"6 weeks, 14 days"**	

```
Dim diff As TimeSpan = endDate.Subtract(startDate)
lbTimeSpan.Text = diff.Humanize(2, App.Settings.CultureInfo)
```
I would have expected 1 months, x weeks, x days, or in this case 8 weeks.

Is this a bug or a wrong usage of the Humanize function?
I'm using v2.2. from NuGet

schnitty:
@MehdiK As requested in Issue #651 I have submitted the work I have done trying to fix the issue.
This PR is not ready to be merged as its completion would likely result in a breaking change.
The current implementation is partial and requires decisions by main contributors on how to take further.
I have written some tests but have commented them out as they fail but can be used when trying to correct the issues.

### DIFF
--- a/src/Humanizer.Tests.Shared/TimeSpanHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/TimeSpanHumanizeTests.cs
@@ -72,6 +72,70 @@ namespace Humanizer.Tests
             Assert.Equal(expected, actual);
         }
 
+        //[Theory]
+        //[InlineData(7, "1 week")]
+        //[InlineData(8, "1 week, 1 day")]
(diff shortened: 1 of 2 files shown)


onovotny:
Closing due to age. Please submit a new PR for the issue when ready.

PR #696, opened 2018-03-16
Improve german number to word

AndreasAdler:
* Add support for trillions, quadrillions and quintillions
* Add special handling for Swiss and Liechtenstein spelling rules
* Add grammatical gender handling
* Improve implementation by adding methods for common code structures

 - [x] Implementation is clean
 - [x] No ReSharper warnings


NuSpecs/Humanizer.Core.de-CH.nuspec
```diff
@@ -0,0 +1,23 @@
...
+    <description>Humanizer Locale (de-CH)</description>
+    <copyright>Copyright 2012-2015  Mehdi Khalili</copyright>
+    <licenseUrl>https://raw.githubusercontent.com/Humanizr/Humanizer/master/LICENSE</licenseUrl>
+    <language>de-CH</language>
+    <dependencies>
+       <dependency id="Humanizer.Core" version="[$version$]" />
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Humanizer\bin\Release\netstandard1.0\de-CH\*.*" target="lib\netstandard1.0\de" />
```
hangy: I think this needs to be in the `de-CH` subdirectory, see https://github.com/Humanizr/Humanizer/blob/24c3a83b744a8d1d41e0d68030ac4dbafc1bea05/NuSpecs/Humanizer.Core.bn-BD.nuspec#L21

NuSpecs/Humanizer.Core.de-LI.nuspec
```diff
@@ -0,0 +1,23 @@
...
+    <description>Humanizer Locale (de-LI)</description>
+    <copyright>Copyright 2012-2015  Mehdi Khalili</copyright>
+    <licenseUrl>https://raw.githubusercontent.com/Humanizr/Humanizer/master/LICENSE</licenseUrl>
+    <language>de-LI</language>
+    <dependencies>
+       <dependency id="Humanizer.Core" version="[$version$]" />
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Humanizer\bin\Release\netstandard1.0\de-LI\*.*" target="lib\netstandard1.0\de" />
```
hangy: I think this needs to be in the `de-LI` subdirectory, see https://github.com/Humanizr/Humanizer/blob/24c3a83b744a8d1d41e0d68030ac4dbafc1bea05/NuSpecs/Humanizer.Core.bn-BD.nuspec#L21

src/Humanizer.Tests.Shared/Localisation/de/NumberToWordsTests.cs
```diff
@@ -7,7 +7,7 @@ public class NumberToWordsTests
     {
         [Theory]
         [InlineData(0, "null")]
-        [InlineData(1, "eins")]
+        [InlineData(1, "ein")]
```
hangy: Would this make `1.ToWords() == "ein"`? I agree that ie. "1 Haus" should be "Ein Haus", but when displayed ie. in a table, `1` should be [`"Eins"`](https://de.wikipedia.org/wiki/Eins), unless I'm missing something.

NuSpecs/Humanizer.Core.de-CH.nuspec
AndreasAdler: Yep, I overlooked that...

NuSpecs/Humanizer.Core.de-LI.nuspec
AndreasAdler: Overlooked that one, too.

src/Humanizer.Tests.Shared/Localisation/de/NumberToWordsTests.cs
AndreasAdler: Yes you're right, this is in a sense a "breaking" change. However, since `ToQuantity` uses `ToWords` when called with `ShowQuantityAs.Numeric` the current output of `"eins"` is grammatically wrong in any case because it's the numbers name when counting.
As far as I see there is no possibilty to add this corner case with the current architecture of Humanizer.

I don't know about other languages but for German we would need for possible outputs for the number `1`:
* `"ein"` for `GrammaticalGender.Masculine`,
* `"eine"` for `GrammaticalGender.Feminine`,
* `"ein"` for `GrammaticalGender.Neuter` and
* `"eins"` when you just need the number's name (something like `GrammaticalGender.None` if you will)

I'm aware this is a limitation but as said I think from the grammatical perspective this implementation is more correct since the method always gets a `GrammaticalGender` provided (by default `GrammaticalGender.Masculine`) and in this case the output `"eins"` is wrong.

hangy: Yup I guess that's a limitation in Humanizer right now. I just wanted to make sure this change was made deliberately. Other than that, I think this PR is a great improvement. 👍

PR #716, opened 2018-06-09
Add heading humanization

corstian:
See the changed readme for more information about the functionality added with this pull request. I would love to implement string heading to numeric heading but I need to figure out how to fetch the localized strings and their keys from the resource files. If anyone has ideas for this, hit me up! :)

 - [X] Implementation is clean
 - [X] No ReSharper warnings


corstian:
@onovotny This branch got out of date with current master. Shall I merge the latest changes into it? Have any other feedback on this PR?

src/Humanizer.Tests/PublicApiApprovalTest.approve_public_api.approved.txt
```diff
@@ -164,6 +164,12 @@ namespace Humanizer
         Feminine = 1,
         Neuter = 2,
     }
+    public class static HeadingExtensions
+    {
+        public static double FromShortHeading(this string heading) { }
+        public static string ToHeading(this double heading, bool shortResult = False, System.Globalization.CultureInfo culture = null) { }
```
onovotny: I would prefer to see an Enum here instead of a boolean since I believe it makes the intent more clear.

I can have two values and default the same.

src/Humanizer/Properties/Resources.resx
```diff
@@ -579,4 +579,100 @@
...
+    <value>west-northwest</value>
+  </data>
+  <data name="NW" xml:space="preserve">
+    <value>northwest</value>
+  </data>
+  <data name="NNW" xml:space="preserve">
+    <value>north-northwest</value>
+  </data>
+  <data name="N_Short" xml:space="preserve">
+    <value>n</value>
```
onovotny: Should short abbreviations be caps?

src/Humanizer.Tests/PublicApiApprovalTest.approve_public_api.approved.txt
corstian: Hmm that makes more sense indeed. What do you mean with the two values thing?

src/Humanizer/Properties/Resources.resx
corstian: Because all other string values are lowercase I chose to do these in lowercase, too. Not sure what would be preferred in daily use cases though.

Issue #718, opened 2018-06-21
Add explicit option to return "0 seconds" instead of "no time"


icnocop:
Hi.

I'd like the ability to explicitly instruct Humanizer to return "0 seconds" instead of "no time".

Thank you.

**Edit**:
For example,
```
TimeSpan.Zero.Humanize(1, toWords: true)); // returns "no time"
TimeSpan.Zero.Humanize(1, toWords: false)); // also returns "no time", but maybe this should return "0 seconds" instead?
```

wcsanders1:
To clarify, does this issue entail making it so that when `toWords` is `false`, the method **always** returns `0 seconds` (or `0 minutes`, `0 hours`, etc., depending on what the minimum time unit is), when the time span is zero?

icnocop:
Hi.

Yes, that is correct.

Current behavior:
```
TimeSpan.Zero.Humanize(1, toWords: true)); // returns "no time"
TimeSpan.Zero.Humanize(1, toWords: false)); // returns "no time"
```
New requested behavior:
```
TimeSpan.Zero.Humanize(1, toWords: true)); // returns "no time"
TimeSpan.Zero.Humanize(1, toWords: false)); // returns "0 seconds"
TimeSpan.Zero.Humanize(1, minUnit: TimeUnit.Day, toWords: false)); // returns "0 days"
```

This request is flexible so if there's a better strategy, I'm all for it, :)

Thank you.

wcsanders1:
I just made a PR that should resolve this issue.

Ashera138:
It looks like this can be closed now. The [PR for this issue](https://github.com/Humanizr/Humanizer/pull/732) was merged. 👍

icnocop:
Thank you! :smile:

onovotny:
2.5.1 is live on NuGet now.

PR #732, opened 2018-07-21
Fixes issue #718

wcsanders1:
 - [x] Implementation is clean
 - [x] No ReSharper warnings


wcsanders1:
I think the changes in this PR will resolve issue 718. Let me know if there's anything else I need to do with regard to this PR.

src/Humanizer.Tests/Humanizer.Tests.csproj
```diff
@@ -24,9 +24,10 @@
     </Compile>
   </ItemGroup>
   <ItemGroup>
-    <Content Include="ApiApprover\PublicApiApprovalTest.approve_public_api.approved.txt">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-    </Content>
+    <ContentWithTargetPath Include="ApiApprover\PublicApiApprovalTest.approve_public_api.approved.txt">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
onovotny: This should be PreserveNewest or it breaks incremental builds. Why was this changed?

src/Humanizer.Tests/Humanizer.Tests.csproj
```diff
@@ -24,9 +24,10 @@
...
+      <TargetPath>PublicApiApprovalTest.approve_public_api.approved.txt</TargetPath>
```
onovotny: Why did this need to change?

src/Humanizer.Tests/Humanizer.Tests.csproj
wcsanders1: I made this change because the ApiApproved test was failing and I thought that this change would fix it. I was wrong. Reverted in e968e34.

src/Humanizer.Tests/Humanizer.Tests.csproj
wcsanders1: The ApiApprover test was failing and it appeared to me that the reason for this was that the approved.txt file was not being copied to the directory where the test runner expected it to be. I made this change, which appeared to solve the problem. However, after reverting the change, the test still passes, so I guess the change was unnecessary. Reverted in 947466a.

onovotny:
Please update the `readme.md` with the new functionality and this will be good to merge. Thanks!

wcsanders1:
Readme updated in 54a060e.

PR #741, opened 2018-08-21
dev/Fixed ToTitleCase bug - issue#607

### LINKED ISSUE #607: Titleize Casing doesn't apply to first word within a bracket
If the following sentence was run through the Titleize string transformer "my summer vacation (with my family)", it would come out as "My Summer Vacation (with My Family)". Ignoring that the word "my" could start with a lower or upper case M, the "with" is not modified at all.

In some cases, the current output is as expected (EG: "Madonna (written by E. Cales in 1998)"), but in other cases the output is unexpected (My above example).

Is this the intended functionality?

Ashera138:
- [x] Implementation is clean
 - [x] No ReSharper warnings

Closes #607


Ashera138:
I just noticed that the `TransformersTests` class. If I need to move the unit test coverage there instead of in `Humanizer.Tests.CasingTests`, let me know. 😃

Issue #719, opened 2018-06-21
Add option to return full words instead of abbreviations for ByteSize objects.


icnocop:
I'd like the ability to explicitly return full words instead of abbreviations for ByteSize objects.

For example,
Current behavior:
```
7.Bits().ToString();         // 7 b
8.Bits().ToString();         // 1 B
```
Feature request:
```
7.Bits().Humanize().ToFullWords(); // 7 bits
8.Bits().ToFullWords();            // 1 Byte
```
Or maybe we can pass configuration options like suggested in https://github.com/Humanizr/Humanizer/issues/508
```
7.Bits().Humanize(new HumanizeOptions{ FullWords = true }); // 7 bits
8.Bits().Humanize(new HumanizeOptions{ FullWords = true }); // 1 Byte
```

Thank you.

Ashera138:
I'd like to give this a shot. 👍 I'll let you know if I have any questions or when I have a pr ready if that's alright with you.

Ashera138:
I'm heading out of town for a couple of days, but I'll pick this back up when I get back. :)

Ashera138:
Alright, PR is ready. Let me know if there are any changes that need to be made! 😃

Ashera138:
Alright, PR is submitted. Let me know if there are any changes that need to be made! 😃

Ashera138:
This can be closed now. 🏷

PR #742, opened 2018-08-29
Fixes Issue: #719 Added ByteSize.ToFullWords() to return full words instead of abbreviations

Ashera138:
 - [x] Implementation is clean
 - [x] No ReSharper warnings
 - [] Run either `build.cmd` or `build.ps1` and ensure there are no test failures
Fixes issue #719.

I'm getting a failing test when running build.cmd, but it's unrelated to any code I changed. The picture below shows this:

testfailure

Please let me know if I need to make any changes. Thank you very much! 😄


onovotny:
This looks good, but my only question is if the words should be upper-case? I think it should be something like 10 megabytes, not 10 Megabytes?

Also, how does it handle singular vs plural? 1 megabyte vs 10 megabytes? It's not as much of an issue with abbreviations, but I think would stick out when expanded to words?

readme.md
```diff
@@ -903,7 +903,7 @@ Also the reverse operation using the `FromMetric` extension.
 ```
 
 ### <a id="bytesize">ByteSize</a>
-Humanizer includes a port of the brilliant [ByteSize](https://github.com/omar/ByteSize) library.
+Humanizer includes a part of the brilliant [ByteSize](https://github.com/omar/ByteSize) library.
```
hangy: "port" should be correct IMHO

hangy:
> This looks good, but my only question is if the words should be upper-case? I think it should be something like 10 megabytes, not 10 Megabytes?

That should depend on the locale. In English, they would use "1 byte", but in German, we use "1 Byte".

onovotny:
@hangy question is what the default should be? There are already ways to lower case or TitleCase things. Or if there are known rules, then the implementation could do the right thing based on the locale. I don't know what all the rules are for when it should start upper or not. Is there a list somewhere?

Ashera138:
I'm seeing both uppercase and lowercase used online, but lowercase seems to be more common. I can't find any source that defines actual capitalization standards, but pages such as [these](https://techterms.com/help/data_storage_units_of_measurement) tend to show the words in lowercase.

Also, let me know if you would like `part` changed back to `port`.

PR #754, opened 2018-10-03
Updated Regex pattern in ToTitleCase method to handle apostrophes

### LINKED ISSUE #584: Transform(To.TitleCase) inconsistent transformations(?)
I am seeing inconsistent transformations when using `To.TitleCase` with all capital strings, but if I transform them to lower case and than use title casing the results are correct:

```
"SUICIDE SQUAD".Transform(To.TitleCase)
"SUICIDE SQUAD".Transform(To.LowerCase).Transform(To.TitleCase)
```

No Change:

> SUICIDE SQUAD
> Suicide Squad

```
"PETE'S DRAGON".Transform(To.TitleCase)
"PETE'S DRAGON".Transform(To.LowerCase).Transform(To.TitleCase)
```

Transform stops after first hyphenated word:

> Pete's DRAGON
> Pete's Dragon

```
"CAPTAIN AMERICA: CIVIL WAR".Transform(To.TitleCase)
"CAPTAIN AMERICA: CIVIL WAR".Transform(To.LowerCase).Transform(To.TitleCase)
```

Only second word is transformed:

> CAPTAIN America: CIVIL WAR
> Captain America: Civil War

Version:

> package id="Humanizer" version="2.1.0" targetFramework="net45"

Ashera138:
I updated the Regex pattern in the `ToTitleCase` method to handle words with apostrophes, as mentioned in issue #584.

 - [x] Implementation is clean
 - [x] No ReSharper warnings

Fixes #584.

Side note, the code as is can't differentiate between contractions and names like "O'Connell". So for example, if you write code like this:

    "Mc'Donald".Transform(To.LowerCase).Transform(To.TitleCase)
it will return `Mc'donald` instead of `Mc'Donald` like some people might expect. That's probably a pretty uncommon case, but it's worth mentioning.

### DIFF
--- a/src/Humanizer.Tests.Shared/TransformersTests.cs
+++ b/src/Humanizer.Tests.Shared/TransformersTests.cs
@@ -1,4 +1,4 @@
-using Xunit;
+﻿using Xunit;
 
 namespace Humanizer.Tests
 {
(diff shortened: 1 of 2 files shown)


Ashera138:
Ah I need to make another change... I'll submit it in a minute here.

Ashera138:
Alright, I'm happy with these changes now. Let me know what you think about this fix and whether or not you want to add extra logic to deal with names that contain apostrophes like I described above.

onovotny:
Thanks!

PR #775, opened 2019-01-10
fixes #526

### LINKED ISSUE #526: Request: Measure of length
I think it would make a good addition to Humanizer if it was possible to humanize length just like time (as with TimeSpan).

Example of how the API could look like. I guess there are more to this feature but it is a start:
```
var value = 1280; // millimeters
value.ToLength(precision: 1, fromUnit: LengthUnit.Millimeters, minUnit: LengthUnit.Meter, maxUnit: LengthUnit.Meter);
```

`value` would become `"1.3 m"`

JonasBarka:
 - [x] Implementation is clean
 - [ ] No ReSharper warnings **Do not got ReSharper.**
 - [x ] There is proper unit test coverage
 - [x] If the code is copied from StackOverflow (or a blog or OSS) full disclosure is included. That includes required license files and/or file headers explaining where the code came from with proper attribution **NA**
 - [x] Your PR is (re)based on top of the latest commits from the `dev` branch (more info below). **Based on latest master.**

Note the interaction between parameters `decimals` and `minPrefix` as described in the comment to one of the unit tests.

Only `maxPrefix` and `minPrefix` are new options, but I added all options to the readme description of `ToMetric`.


src/Humanizer/MetricNumeralExtensions.cs
```diff
@@ -1,8 +1,8 @@
-﻿// Wrote by Alois de Gouvello https://github.com/aloisdg
+﻿// Written by Alois de Gouvello https://github.com/aloisdg with additions by Jonas Barkå https://github.com/JonasBarka.
```
onovotny: The copyrights at the top of the file, if any, should be simply `Copyright (c) .NET Foundation and Contributors`

Alternatively, it can be removed as it's in the project root.

src/Humanizer/MetricNumeralExtensions.cs
```diff
@@ -1,8 +1,8 @@
...
 
 // The MIT License (MIT)
 
-// Copyright (c) 2015 Alois de Gouvello
+// Copyright (c) 2015-2019 Alois de Gouvello and each other contributor for their respective work.
```
onovotny: Same as above

SimonCropp:
@JonasBarka do you still want this one merged? if so can you address the comments and rebase

PR #787, opened 2019-02-19
fixes #772

### LINKED ISSUE #772: ToMetric() - hasSpace behaviour with no exponent intended behavior?
ToMetric has an optional parameter *hasSpace* for inserting a space before the exponent.

>1000.ToMetric(hasSpace: true, ) + "m" => "1 km"
>0.001.ToMetric(hasSpace: true, ) + "m" => "1 km"

But the space is not inserted at the end when no prefix is added.

>10.ToMetric(hasSpace:true, ) + "m" => "10m"

Indented behavior? If the main intended usage of ToMetric is to also add a unit, as the name of the method implies, the this seems like a bug. But in isolation not adding it would be more correct.

An option for always adding the space would probably be best, or what do you think?

defigli:
 - [x] Implementation is clean
 - [x] No ReSharper warnings
 - [ ] Your PR is (re)based on top of the latest commits from the `dev` branch (more info below)


onovotny:
Please resolve the conflicts and then this PR can be merged. Thanks

PR #793, opened 2019-03-04
Change Pascalize to remove spaces fix #774

### LINKED ISSUE #774: Pascalize and Camelize do not join words separated by spaces
My expectation of these methods would be that it would join separate words together.

`"This is a sentence".Pascalize();`

Expected result: `"ThisIsASentence"`
Actual result: `"This is a sentence"`

My workaround for now i to `Underscore` first:
`"This is a sentence".Underscore().Pascalize()`

AKTheKnight:
This is a PR to change to fix #774 that changes Pascalize (and therefore Camelize too) to treat spaces in the same way it treats underscores.

I wasn't sure whether to have this as an optional change, or a change to the method itself, so I've chosen to just change the method and create this PR as a way to generate conversation around the issue.

Let me know if you have any thoughts.

### DIFF
--- a/src/Humanizer/InflectorExtensions.cs
+++ b/src/Humanizer/InflectorExtensions.cs
@@ -72,7 +72,7 @@ namespace Humanizer
         public static string Pascalize(this string input)
         {
             
-            return Regex.Replace(input, "(?:^|_| )(.)", match => match.Groups[1].Value.ToUpper());
+            return Regex.Replace(input, "(?:^|_| +)(.)", match => match.Groups[1].Value.ToUpper());
             
             //return Regex.Replace(input, @"\s+", "");
         }
--- a/src/Humanizer/InflectorExtensions.cs
+++ b/src/Humanizer/InflectorExtensions.cs
@@ -70,11 +70,8 @@ namespace Humanizer
         /// <param name="removeWhiteSpace">"This has a space" -> "ThisHasASpace" if true</param>
         /// <returns></returns>
         public static string Pascalize(this string input)
-        {
-            
+        {   
             return Regex.Replace(input, "(?:^|_| +)(.)", match => match.Groups[1].Value.ToUpper());
-            
-            //return Regex.Replace(input, @"\s+", "");
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


onovotny:
Overall this change is fine, but something seems to have gotten messed up with the readme diff?

AKTheKnight:
Not sure what happened there, think I've fixed it and updated the branch

src/Humanizer/InflectorExtensions.cs
```diff
@@ -67,20 +67,25 @@ public static string Titleize(this string input)
         /// By default, pascalize converts strings to UpperCamelCase also removing underscores
         /// </summary>
         /// <param name="input"></param>
+        /// <param name="removeWhiteSpace">"This has a space" -> "ThisHasASpace" if true</param>
         /// <returns></returns>
         public static string Pascalize(this string input)
         {
-            return Regex.Replace(input, "(?:^|_)(.)", match => match.Groups[1].Value.ToUpper());
+            
+            input = Regex.Replace(input, "(?:^|_| )(.)", match => match.Groups[1].Value.ToUpper());
```
onovotny: RegEx isn't my strong suit, but is there a way these two regex statements can be combined into one?

AKTheKnight: I'll have a look what I can come up with

AKTheKnight:
The latest commits remove the second regex line (it wasn't actually needed), and also add support for multiple spaces in a row.

Not sure what the thoughts are on adding support for multiple underscores in a row?

src/Humanizer/InflectorExtensions.cs
```diff
@@ -67,20 +67,25 @@ public static string Titleize(this string input)
...
         /// <param name="input"></param>
+        /// <param name="removeWhiteSpace">"This has a space" -> "ThisHasASpace" if true</param>
         /// <returns></returns>
         public static string Pascalize(this string input)
         {
-            return Regex.Replace(input, "(?:^|_)(.)", match => match.Groups[1].Value.ToUpper());
+            
+            return Regex.Replace(input, "(?:^|_| +)(.)", match => match.Groups[1].Value.ToUpper());
+            
+            //return Regex.Replace(input, @"\s+", "");
```
onovotny: Please remove the old code and extra spaces in this method.

PR #800, opened 2019-03-25
[WIP] Fix #651

### LINKED ISSUE #651: Timespan.humanize() shows weird results

AKTheKnight:
This is a working PR to look at fixing the issues described in #651.

Currently it just includes the changes from the closed PR #681 as a baseline.

Tests will fail, we need to work out how we want to deal with certain situations, as I find them I will comment and see how people want to go forward

### DIFF
--- a/src/Humanizer.Tests.Shared/TimeSpanHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/TimeSpanHumanizeTests.cs
@@ -71,6 +71,69 @@ namespace Humanizer.Tests
             var actual = TimeSpan.FromDays(days).Humanize();
             Assert.Equal(expected, actual);
         }
+        
+        [Theory]
+        [InlineData(7, "1 week")]
+        [InlineData(8, "1 week, 1 day")]
+        [InlineData(9, "1 week, 2 days")]
+        [InlineData(14, "2 weeks")]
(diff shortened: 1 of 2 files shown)


AKTheKnight:
The most common failure is that a time of zero now returns "no time", and not "0 [units]" as before. What are peoples thoughts about it?

It does mean we no longer respect the "minUnit" parameter by not returning "0 [units]", not sure how people feel about that.

AKTheKnight:
A lot of the rest of the failing errors also look to be related to how we handle months, with this method a month is 30.436.... days. Not sure how we want to solve that

PR #803, opened 2019-03-26
fix: add ByteRate formatting and value semantics

### LINKED ISSUE #802: ByteRate has no ToString or comparison methods.
ByteRate, unlike ByteSize has no ToString or IComparable<>/IEquatable<> methods.
I'm using the project in some PowerShell cmdlets and the lack of ToString means that any ByteRate object is written to the console as just `Humanizer.Bytes.ByteRate`

Cirzen:
## Summary

`ByteRate` values now render as human-readable transfer rates by default, so PowerShell and console output show values such as `400 B/s` instead of the type name. Rates can also be compared, sorted, and tested for equality after normalization to bytes per second, with hashing consistent with equality. The existing `Humanize` formatting API remains unchanged.

Fixes #802

## Validation

- `dotnet test` passed on .NET 8, 10, and 11: 57,017 tests per target
- Release package build completed without warnings or errors
- `dotnet format Humanizer.slnx --verify-no-changes` passed
- CodeRabbit approved the final head and all CI checks passed

- [x] Implementation is clean
- [x] Code adheres to the existing coding standards
- [x] No code was copied from StackOverflow, a blog, or another OSS project
- [x] There are very few or no comments
- [x] XML documentation is added/updated for the addition/change
- [x] The PR is updated on top of the latest `main` branch
- [x] The issue being fixed is linked with `Fixes #802`
- [x] The existing ByteRate documentation remains accurate
- [x] Current supported-target test, pack, and format workflows pass


Cirzen:
Looks like my typo correction has caused one of the tests to fail. I'll revert that change.

Cirzen:
Failing on the approve_public_api test, since that's what I've changed.
Was 
```
    public class ByteRate
    {
        public ByteRate(Humanizer.Bytes.ByteSize size, System.TimeSpan interval) { }
        public System.TimeSpan Interval { get; }
        public Humanizer.Bytes.ByteSize Size { get; }
        public string Humanize(Humanizer.Localisation.TimeUnit timeUnit = 1) { }
        public string Humanize(string format, Humanizer.Localisation.TimeUnit timeUnit = 1) { }
    }
```
Now
```
    public class ByteRate : System.IComparable, System.IComparable<Humanizer.Bytes.ByteRate>, System.IEquatable<Humanizer.Bytes.ByteRate>
    {
        public ByteRate(Humanizer.Bytes.ByteSize size, System.TimeSpan interval) { }
        public System.TimeSpan Interval { get; }
        public Humanizer.Bytes.ByteSize Size { get; }
        public int CompareTo(Humanizer.Bytes.ByteRate other) { }
        public int CompareTo(object obj) { }
        public bool Equals(Humanizer.Bytes.ByteRate other) { }
        public string Humanize(Humanizer.Localisation.TimeUnit timeUnit = 1) { }
        public string Humanize(string format, Humanizer.Localisation.TimeUnit timeUnit = 1) { }
        public override string ToString() { }
        public string ToString(string format, Humanizer.Localisation.TimeUnit timeUnit = 1) { }
    }
```
Could do with some feedback or guidance on best way to proceed here.

AKTheKnight:
If you run either build.cmd or build.ps1 it should come up with a method of updating the PublicApiApprovalTest.approve_public_api.approved.txt file.

Once that matches the public api, the test will then pass

src/Humanizer/Bytes/ByteRate.cs
```diff
@@ -76,5 +76,58 @@ public string Humanize(string format, TimeUnit timeUnit = TimeUnit.Second)
...
+        public bool Equals(ByteRate other)
+        {
+            return CompareTo(other) == 0;
+        }
+
+        /// <inheritdoc />
+        public int CompareTo(Object obj)
+        {
+            if (obj == null) return 1;
+            var cmp = (ByteRate)obj;
```
onovotny: This method should call the other overload of `CompareTo` with the `ByteRange` -- `CompareTo(obj as ByteRange)`

src/Humanizer/Bytes/ByteRate.cs
```diff
@@ -76,5 +76,58 @@ public string Humanize(string format, TimeUnit timeUnit = TimeUnit.Second)
...
+
+        /// <inheritdoc />
+        /// <summary>
+        /// Checks if two ByteRate objects have the same equivalent rate
+        /// </summary>
+        /// <param name="other"></param>
+        /// <returns>True if rates are equivalent, otherwise false</returns>
+        public bool Equals(ByteRate other)
+        {
+            return CompareTo(other) == 0;
```
onovotny: Is there a faster/more direct equality check than through the compare?

Cirzen: Not that I can think of without either: duplicating code, or losing the sense of equivalence that the final rate is the important factor, not the given ByteSize and Interval values. But perhaps a small amount of code duplication is OK to keep the IEquatable<T> and IComparable<T> methods not dependent on each other...
```suggestion
            return (Size.Bytes / Interval.TotalSeconds) - (other.Size.Bytes / other.Interval.TotalSeconds) == 0;
```

Cirzen:
Hi Claire, would you be willing to have another look at this, I've hopefully resolved any remaining concerns.

clairernovotny:
Looks like there's a build error that needs to be resolved?

Cirzen:
You're right, I jumped the gun. Looks like some issues in Master

Cirzen:
Now passing tests

SimonCropp:
@Cirzen  sorry for the delay. can you rebase and re-approve the api changes

clairernovotny:
> can you rebase and re-approve the api changes

Updated in 57f347fe via a non-force merge from current main. The change now adds ByteRate.ToString() and normalized IComparable support, with focused tests and all four current public-API snapshots updated. The earlier IEquatable implementation was not retained because it did not provide a coherent Equals(object)/GetHashCode contract.

Validated on .NET 8, 10, and 11 (57,015 tests per target), plus Release pack and dotnet format verification.

src/Humanizer/Bytes/ByteRate.cs
```diff
@@ -6,7 +6,9 @@ namespace Humanizer;
 /// <remarks>
 /// Create a ByteRate with given quantity of bytes across an interval
 /// </remarks>
-public class ByteRate(ByteSize size, TimeSpan interval)
+public class ByteRate(ByteSize size, TimeSpan interval) :
+    IComparable<ByteRate>,
+    IComparable
```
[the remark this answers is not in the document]
clairernovotny: > Add the missing `ByteRate` equality contract.

Addressed in 418c2b00: equality and hashing now use the normalized bytes-per-second value, with API snapshots and focused equality-contract tests.

PR #807, opened 2019-04-01
[WIP] Fix #765. Change strategy to add Weeks. Add Week property

### LINKED ISSUE #765: DateTime Humanizer : returns hours, days, months, but not weeks
Why does DateTime Humanizer return a number of hours, days, months, but never return a number of weeks ?
It does not seems consistent with TimeSpan Humanizer.

Example :
TimeSpan.FromDays(7).Humanize() => "1 week"
DateTime.UtcNow.AddDays(7).Humanize() => "7 days from now" (instead of "1 week from now")

AKTheKnight:
This also breaks lots of tests that are missing the property, I'll work on fixing them

### DIFF
--- a/src/Humanizer/DateTimeHumanizeStrategy/DateTimeHumanizeAlgorithms.cs
+++ b/src/Humanizer/DateTimeHumanizeStrategy/DateTimeHumanizeAlgorithms.cs
@@ -150,9 +150,14 @@ namespace Humanizer.DateTimeHumanizeStrategy
                 return formatter.DateHumanize(TimeUnit.Day, tense, days);
             }
 
-            if (ts.TotalDays < 28)
+            if (ts.TotalDays < 7)
             {
                 return formatter.DateHumanize(TimeUnit.Day, tense, ts.Days);
+            }
+
+            if (ts.TotalDays < 28)
+            {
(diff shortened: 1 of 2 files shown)


src/Humanizer.Tests.Shared/DateHumanizeDefaultStrategyTests.cs
```diff
@@ -88,8 +88,8 @@ public void HoursFromNowNotTomorrow(int hours, string expected)
 
         [Theory]
         [InlineData(1, "yesterday")]
-        [InlineData(10, "10 days ago")]
-        [InlineData(27, "27 days ago")]
```
SimonCropp, line 92: given this test passes in TimeUnit.Day, why would it result in week/weeks

also 10 days is no "on week"

AKTheKnight: Good point, I think this is going to be a culture thing and then a question of how precise people would like humanizer to be.

7 days would be "a week ago", and in fact if someone asked me about something 8 days ago I might also accept "a week ago".

Perhaps I should look at making this only provide weeks if the value is one of 7,14,21,28. And then fall backs to months as expected after that 🤔

PR #826, opened 2019-06-09
add BolgarianNumberToWordsConverter

oxi-mix:
added partial BolgarianNumberToWordsConverter
added tests for them

### DIFF
--- a/src/Humanizer.Tests.Shared/Localisation/bg/NumberToWordsTests.cs
+++ b/src/Humanizer.Tests.Shared/Localisation/bg/NumberToWordsTests.cs
@@ -1,0 +1,161 @@
+﻿using Xunit;
+
+namespace Humanizer.Tests.Localisation.bg
+{
+    [UseCulture("bg")]
+    public class NumberToWordsTests
+    {
+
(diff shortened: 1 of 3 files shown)


AKTheKnight:
Not your fault but looking at this PR has made me spot a few issues with other aspects of the codebase:
The ToWords extensions takes a long (this) and a CultureInfo.
But, the implementations I have looked at use the below code:

```csharp
if (input > Int32.MaxValue || input < Int32.MinValue)
{
        throw new NotImplementedException();
}
```
The throwing a NotImplementedException is great, but I'm sure this must have been incredibly frustrating to people at some point or another trying to use this.

I'm not sure where we go from here @onovotny, are you happy to accept this PR and then I could look at working out what would be needed to fix all the exceptions, or would you like me to work with oxi-mix on fixing this one before merging?

onovotny:
Thanks!

I'm happy to look at ways where the exceptions wouldn't be needed.

oxi-mix:
Let me ask you a question? When will these changes be released? We really need these changes in our project.

PR #849, opened 2019-09-05
Implement Portuguese ToWords & ToOrdinalWords

hazzik:
According to Wikipedia all lusophone (Portuguese speaking) countries except Brazil use long scale.[<sup>[1]</sup>](https://pt.wikipedia.org/wiki/Escalas_curta_e_longa)

As far as it concerns Humanizer the only difference is 10<sup>9</sup> which in Brazil would be "bilhão" and in rest of Portuguese speaking countries it would be "mill milhões". For ordinal it is "bilionésimo/a" and "milésimo/a milionésimo/a"[<sup>[2]</sup>](https://ciberduvidas.iscte-iul.pt/consultorio/perguntas/milesimo-milionesimo/31914) respectively.


onovotny:
Looks good! @hazzik Can you please sign the CLA bot and then I can merge?

PR #851, opened 2019-09-09
remove special case for 0 to fix "hasSpace: true"

davepermen:
See: https://dotnetfiddle.net/NHTOam

Console.WriteLine(0.ToMetric(hasSpace: true) + "Wh");
Console.WriteLine(1.ToMetric(hasSpace: true) + "Wh");
Console.WriteLine(1000.ToMetric(hasSpace: true) + "Wh");

outputs

0Wh
1 Wh
1 kWh


onovotny:
Can you please add/update the tests to show the expected/correct behavior?

PR #875, opened 2019-10-31
Add a time expression provider to allow the user to select the predefined ones.

### LINKED ISSUE #816: Support for time expressions ("from now", "in", "since", "ago")
Hi,

Thank you for this great library ! Such a usefull tool !

I was thinking maybe it will be a great add to be able to choose between "since" and "ago" in order to express ifferents meanings for past expression and be able to choose between "from now" and "in" in order to express differents meanings for future expression.

- since express that whatever it is it is what it is since X time (ex: since one year)
- ago express that whatever it is it was what it was X time ago (ex: one year ago) (and may change since then)
- from now express that whatever it is it will be what it is X time from now (ex: 1 hour from now)
- in express that whatever it is it will be what it will be in X time (ex: in 1 hour) (and may be different before)

Today there is no options it's "ago" and that's all:
`DateTime.UtcNow.AddHours(-30).Humanize() => "yesterday"`
`DateTime.UtcNow.AddHours(-2).Humanize() => "2 hours ago"`

But maybe it can be more with a choice between "since" and "ago":
`DateTime.UtcNow.AddHours(-30).Humanize(PastTimeExpression.Ago) => "yesterday"`
`DateTime.UtcNow.AddHours(-2).Humanize(PastTimeExpression.Ago) => "2 hours ago"`
`DateTime.UtcNow.AddHours(-30).Humanize(PastTimeExpression.Since) => "since yesterday"`
`DateTime.UtcNow.AddHours(-2).Humanize(PastTimeExpression.Since) => "since 2 hours"`

Today there is no options it's "from now" and that's all:
`DateTime.UtcNow.AddHours(30).Humanize() => "tomorrow"`
`DateTime.UtcNow.AddHours(2).Humanize() => "2 hours from now"`

"tomorrow" is a specific case because in a way to express "from now" it should be "until tomorrow" because only "tomorrow" is more compatible with a "in" case

But maybe it can be more with a choice between "from now" and "in":
`DateTime.UtcNow.AddHours(30).Humanize(FutureTimeExpression.FromNow) => "until tomorrow"`
`DateTime.UtcNow.AddHours(2).Humanize(FutureTimeExpression.FromNow) => "2 hours from now"`
`DateTime.UtcNow.AddHours(30).Humanize(FutureTimeExpression.In) => "tomorrow"`
`DateTime.UtcNow.AddHours(2).Humanize(FutureTimeExpression.In) => "in 2 hours"`

Some examples :
"Your flight will take off in 2 hours"
"Our plane is flying since 4 hours"
"We will waiting our next flight 3 hours from now"
"We land 1 hour ago"

I think it's not easy to add it but it can be a great add to allow to express those differents meanings with time expressions.

Thank you

luqp:
Fixes #816 

Add support for the `Humanize()` method to work with other time expressions so that the user can select from the predefined.

@MehdiKhalili, @onovotny, @hazzik, @mexx Could you please review this? If the approach is good  I can continue updating the documentation and apply your suggestions.
Thank you.

##
 - [x] Implementation is clean
 - [x] No ReSharper warnings


clairernovotny:
Thank you for exploring configurable time expressions. The resource-key and provider implementation layer used here was replaced by the YAML/source-generator localization architecture in #1688, so this PR no longer applies. The custom-expression idea remains potentially valuable and should be reconsidered against the current formatter design.

Issue #880, opened 2019-11-12
BolgarianNumberToWordsConverter should be renamed to BulgarianNumberToWordsConverter


NikolayIT:
`BolgarianNumberToWordsConverter` should be renamed to `BulgarianNumberToWordsConverter`. The correct naming is Bulgarian.

https://en.wikipedia.org/wiki/Bulgarian_language

AKTheKnight:
This can be closed now :)

hazzik:
Fixed by #882

PR #882, opened 2019-11-16
Rename to BulgarianNumberToWordsConverter.cs

ErwinG01:


AKTheKnight:
Just so it's clear on the issue, fixes #880

PR #888, opened 2019-12-06
Changed bulgarian number to words converter to support numbers up to 1mil

antonmihaylov:
The Bulgarian numbers to words converter was handling numbers up to 1000 before. 
It also didn't handle well round number ordinal conversions (e.g. 1000, 200)

This fixes those problems. Also changed tests to match those changes.


onovotny:
Thanks!

PR #889, opened 2019-12-07
Add czech number to words conversion

Kukkimonsuta:
This PR adds czech support for `ToWords` number conversion. `ToOrdinalWords`  is not implemented (uses default conversion).

There are various ways that some numbers can be read in czech, I've chosen more modern version where gender difference is only for `one` and `two` and not any number ending in `one` or `two`. I've also chosen more explicit version where you say `one million` and not just `million` which should match humanizer behavior in english.


onovotny:
Thanks!

PR #989, opened 2020-10-02
Add number.ToWords() support for large numbers in French

### LINKED ISSUE #988: number.ToWords() should support large numbers in French
It should be possible to display numbers in French up to (short-scale) quintillions, just like in English.

Note that French uses the "long scale" rather than the "short scale". _Refer to [Long and Short Scales Comparison](https://en.wikipedia.org/wiki/Long_and_short_scales#Comparison)._

louis-z:
Fixes #988 

 - [x] Implementation is clean


clairernovotny:
Can we do this without a breaking change? New overloads are fine, but renaming existing methods will break people.

louis-z:
> Can we do this without a breaking change? New overloads are fine, but renaming existing methods will break people.

Absolutely. I will adapt the PR accordingly.

louis-z:
Actually the modified classes are internal. Do you still want me to prevent API breakage?

PR #999, opened 2020-10-23
Non breaking change fixes #781

### LINKED ISSUE #781: ToMetric to million not m
Would it be possible to use to Metric to have a long format as well?

So instead of 1.3M it would say 1.3 Million?

kp73:
Hi, I'd like to add this fix - but I notice an existing pull req. that wasn't merged because it's a breaking change.  

No worries if the existing fix is accepted/in some plan, just wanted to offer this option to get the fix sooner.


clairernovotny:
LGTM.

PR #1005, opened 2020-11-12
Add support for culture-specific ByteSize units

### LINKED ISSUE #810: ByteSize symbols should be translated
Hi,
I'm French, and I'm using Humanizer to display TimeSpans and Byte size in a human readable-format.

However, in France, the term "byte" doesn't exist, except for programmers, and the word "octet" is used instead.

A few examples of english units VS french ones :
- `MB` => `Mo`
- `kB` => `ko`
- `kb` => `kb`

When non-technical folks read "MB", they will read it as "Méga Bits", which is a _bit_ different.

For reference, here is how Windows shows the file sizes:
image

louis-z:
Fixes #810 

 - [x] Implementation is clean


dahlbyk:
FYI: #1062

clairernovotny (approved):

Issue #1013, opened 2020-11-27
Breaking change violating SemVer rules


kellcomnet2:
Please see https://github.com/icsharpcode/ILSpy/issues/2232

The decompiler expects: System.String Humanizer.InflectorExtensions.Singularize(System.String, Boolean)
The newer Humanizer only offers: System.String Humanizer.InflectorExtensions.Singularize(System.String, Boolean, Boolean).

The breaking change is due to the added optional parameter. Note that both are from the same major version, seems like Humanizer does not follow semver rules... not sure how we could fix this.

clairernovotny:
This certainly wasn't intentional. Can you please submit a PR with an overload containing the original parameters and we'll get an update out.

SimonCropp:
Closing due to no response

PR #1015, opened 2020-11-28
Update InflectorExtensions.cs

kellcomnet2:
Fixes Issue #1013


clairernovotny:
The build is still failing here....

kellcomnet2:
Ok that is odd as a simple console app with those as empty methods is building locally.

PR #1025, opened 2021-01-12
Add French support when using toWords: true in TimeSpan humanization

0xced:
Also removed [Trait("Translation", "Google")] from TimeSpanHumanizeTests after  review. (I assumed that the trait means that the translation were automatically generated by Google vs manually written by a human)


aloisdg:
French here. Everything looks good :)

PR #1062, opened 2021-05-02
Parse `ByteSize` with `IFormatProvider`

### LINKED ISSUE #1061: Parse `ByteSize` with `IFormatProvider`?
While poking around for #1060 I noticed `CultureInfo.CurrentCulture` in [`ByteSize.TryParse()`](https://github.com/Humanizr/Humanizer/blob/1285835f92626619ed2d678c8910b1a8a21fcf6e/src/Humanizer/Bytes/ByteSize.cs#L455): https://github.com/Humanizr/Humanizer/blob/1285835f92626619ed2d678c8910b1a8a21fcf6e/src/Humanizer/Bytes/ByteSize.cs#L474

Would you be interested `TryParse`/`Parse` overloads that accept `IFormatProvider`? Asking first since it's a public API change.

It would check if the `IFormatProvider` is either `CultureInfo` or a `NumberFormatInfo`.

dahlbyk:
Closes #1061 

1. Added new overloads for `Parse` and `TryParse`, called with `formatProvider: null` from the existing methods.
2. Using `formatProvider` with `double.TryParse` requires specifying `NumberStyles`, which is more correct anyway (`AllowDecimalPoint`) since there's no handling for negative, currency/exponents don't make sense, etc.
3. Added handling for `DecimalGroupSeparator`, too (e.g. `2,000KB`).
4. Adjusted tests to prove/fix incorrect handling for a missing/unrecognized suffix (e.g.`1000 ` or `1000KBB`).
5. Added missing test for `TryParse(null, ...)`
    - The new test codifies existing behavior; however, it occurs to me that the framework's `TryParse()` methods would just `return false` instead of throwing. Fixing that would be a breaking change, and would motivate additional changes to `Parse()`: given `null`, `throw new ArgumentNullException()`; given empty/whitespace, `throw new FormatException()`. Glad to make adjusments, if you wish.

### Checklist

 - [ ] Implementation is clean
    - Subjective!
 - [ ] ~~If the code is copied from StackOverflow (or a blog or OSS) full disclosure is included. That includes required license files and/or file headers explaining where the code came from with proper attribution~~
    - Unsure what's necessary for additive change to public API. Existing `Parse`/`TryParse` do not have XML docs.
 - [ ] ~~Readme is updated if you change an existing feature or add a new one~~

### DIFF
--- a/src/Humanizer.Tests.Shared/Bytes/ParsingTests.cs
+++ b/src/Humanizer.Tests.Shared/Bytes/ParsingTests.cs
@@ -56,10 +56,13 @@ namespace Humanizer.Tests.Bytes
         [Theory]
         [InlineData("2000.01KB", "")] // Invariant
         [InlineData("2,000.01KB", "")]
+        [InlineData("+2000.01KB", "")]
         [InlineData("2000.01KB", "en")]
         [InlineData("2,000.01KB", "en")]
+        [InlineData("+2000.01KB", "en")]
         [InlineData("2000,01KB", "de")]
         [InlineData("2.000,01KB", "de")]
+        [InlineData("+2000,01KB", "de")]
         public void TryParseWithCultureInfo(string value, string cultureName)
         {
             var culture = new CultureInfo(cultureName);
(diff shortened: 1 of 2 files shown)


clairernovotny:
Looks good, thanks! I'd say that we should fix the behavior around parsing so that it's correct instead of codifying the existing behavior. While technically a breaking change, I'd say it's the right thing to do.

dahlbyk:
> Looks good, thanks! I'd say that we should fix the behavior around parsing so that it's correct instead of codifying the existing behavior. While technically a breaking change, I'd say it's the right thing to do.

Done.

clairernovotny:
Thanks! I can merge once rebased from the other change to bytesize.

dahlbyk:
Seeing #1005 included negative output, I've updated parsing to allow leading signs as well.

clairernovotny:
Thanks!

PR #1063, opened 2021-05-03
Transform with Culture via `ICulturedStringTransformer`

dahlbyk:
As proposed in https://github.com/Humanizr/Humanizer/issues/1061#issuecomment-830826232:

1. Adds support for `input.Transform(culture, transformers)`
2. Adds `ICulturedStringTransformer : IStringTransformer`
   https://github.com/Humanizr/Humanizer/blob/3e23bb41f4a60fcb7933227971672f25bb0df013/src/Humanizer/Transformer/ICulturedStringTransformer.cs#L16
    - New interface avoids a breaking change, but I think it's worth considering just updating `IStringTransformer.Transform()` to accept `culture`. Trivial for custom implementations to be updated with the new parameter, since they can just ignore `culture` if they don't care.
3. Updates `To.LowerCase` and such to return `ICulturedStringTransformer` (shouldn't be a breaking change, since it's assignable to `IStringTransformer`, with internal classes updated to implement the new overload.

### Checklist

 - [x] Implementation is clean
 - [ ] There is proper unit test coverage
    - Haven't figured out how to test, as I don't know a language with upper/lower rules different from English.
 - [ ] ~~If the code is copied from StackOverflow (or a blog or OSS) full disclosure is included. That includes required license files and/or file headers explaining where the code came from with proper attribution~~
 - [ ] ~~Readme is updated if you change an existing feature or add a new one~~

### DIFF
--- a/src/Humanizer/Transformer/ICulturedStringTransformer.cs
+++ b/src/Humanizer/Transformer/ICulturedStringTransformer.cs
@@ -1,0 +1,18 @@
+﻿using System.Globalization;
+
+namespace Humanizer
+{
+    /// <summary>
+    /// Can transform a string with the given culture
+    /// </summary>
+    public interface ICulturedStringTransformer : IStringTransformer
+    {
+        /// <summary>
+        /// Transform the input
+        /// </summary>
+        /// <param name="input">String to be transformed</param>
+        /// <param name="culture">The culture</param>
+        /// <returns></returns>
+        string Transform(string input, CultureInfo culture);
+    }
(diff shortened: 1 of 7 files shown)


dahlbyk:
Just noticed `ApplyCase(LetterCasing)` and `Humanize(LetterCasing)`. Could reasonably update any/all to provide overloads that accept `CultureInfo`, too.

dahlbyk:
Rebased onto `main`. Thoughts?

PR #1068, opened 2021-05-08
Add TimeSpan.ToAge() extension method

### LINKED ISSUE #968: TimeSpan.Humanize should offer an "age" format
Using a TimeSpan to express an age, I would like to output that information in one of 2 ways
- either as an hyphenated adjectival phrase (e.g. a `"10-week-old"` cat), or
- as a non-hyphenated phrase (e.g. the cat is `"10 weeks old"`)

If you feel it would benefit the product, I'd be interested in implementing it.

louis-z:
- Introduce the `"TimeSpanHumanize_Age"` resource, which indicates how to format a humanized TimeSpan as an age expression, when the two differ.
- Add the `Resources.TryGetResource()` method, which tries to retrieve a resource for a given culture but without falling back on the default culture. This will allow us to define the `"TimeSpanHumanize_Age"` resource for `en-US` (aka the default culture), but not have other cultures fall back on that culture's resource.
- Add `DefaultFormatter.TimeSpanHumanize_Age()`, which resorts to `Resources.TryGetResource()` to determine how to express a TimeSpan as age. If no resource is found for the given culture, it will simply output the standard humanized TimeSpan as is. *(For other cultures where standard TimeSpan and age expression differ, such as many germanic languages, `"TimeSpanHumanize_Age"` will eventually need to be defined. However, this is outside the scope of this PR.)*
- Add tests for both English & 1 non-English (i.e. French) cultures

Fixes #968 

 - [x] Implementation is clean

### DIFF
--- a/src/Humanizer.Tests.Shared/Localisation/fr/TimeSpanHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/Localisation/fr/TimeSpanHumanizeTests.cs
@@ -144,6 +144,18 @@
         }
 
         [Theory]
+        [InlineData(4, false, "4 jours")]
+        [InlineData(23, false, "3 semaines")]
+        [InlineData(64, false, "2 mois")]
+        [InlineData(367, true, "un an")]
+        [InlineData(750, true, "deux ans")]
+        public void Age(int days, bool toWords, string expected)
+        {
+            var actual = TimeSpan.FromDays(days).ToAge(toWords: toWords);
+            Assert.Equal(expected, actual);
+        }
+
+        [Theory]
         [InlineData(TimeUnit.Year, "0 an")]
         [InlineData(TimeUnit.Month, "0 mois")]
         [InlineData(TimeUnit.Week, "0 semaine")]
--- a/src/Humanizer/Localisation/Formatters/IFormatter.cs
+++ b/src/Humanizer/Localisation/Formatters/IFormatter.cs
@@ -28,6 +28,13 @@
         string TimeSpanHumanize(TimeUnit timeUnit, int unit, bool toWords = false);
 
         /// <summary>
+        /// Returns the age format that converts a humanized TimeSpan string to an age expression.
+        /// For instance, in English that format adds the " old" suffix, so that "40 years" becomes "40 years old".
+        /// </summary>
+        /// <returns>Age format</returns>
+        string TimeSpanHumanize_Age();
+
+        /// <summary>
         /// Returns the string representation of the provided DataUnit, either as a symbol or full word
         /// </summary>
         /// <param name="dataUnit">Data unit</param>
--- a/src/Humanizer/Localisation/Resources.cs
+++ b/src/Humanizer/Localisation/Resources.cs
@@ -17,5 +17,20 @@ namespace Humanizer
         /// <returns>The value of the resource localized for the specified culture.</returns>
         public static string GetResource(string resourceKey, CultureInfo culture = null) =>
             ResourceManager.GetString(resourceKey, culture);
+
+        /// <summary>
+        /// Tries to get the value of the specified string resource, without fallback
+        /// </summary>
+        /// <param name="resourceKey">The name of the resource to retrieve.</param>
+        /// <param name="culture">The culture of the resource to retrieve. If not specified, current thread's UI culture is used.</param>
+        /// <param name="result">The value of the resource localized for the specified culture if found; null otherwise.</param>
+        /// <returns>true if the specified string resource was found for the given culture; otherwise, false.</returns>
+        public static bool TryGetResource(string resourceKey, CultureInfo culture, out string result)
+        {
+            culture ??= CultureInfo.CurrentUICulture;
+            var resourceSet = ResourceManager.GetResourceSet(culture, createIfNotExists: false, tryParents: false);
+            result = resourceSet?.GetString(resourceKey);
+            return result is not null;
+        }
     }
 }
--- a/src/Humanizer/TimeSpanHumanizeExtensions.cs
+++ b/src/Humanizer/TimeSpanHumanizeExtensions.cs
@@ -44,6 +44,24 @@
             return ConcatenateTimeSpanParts(timeParts, culture, collectionSeparator);
         }
 
+        /// <summary>
+        /// Turns a TimeSpan into an age expression, e.g. "40 years old"
+        /// </summary>
+        /// <param name="timeSpan">Elapsed time</param>
+        /// <param name="culture">Culture to use. If null, current thread's UI culture is used.</param>
+        /// <param name="maxUnit">The maximum unit of time to output. The default value is <see cref="TimeUnit.Year"/>.</param>
+        /// <param name="toWords">Uses words instead of numbers if true. E.g. "forty years old".</param>
+        /// <returns>Age expression in the given culture/language</returns>
+        public static string ToAge(this TimeSpan timeSpan, CultureInfo culture = null, TimeUnit maxUnit = TimeUnit.Year, bool toWords = false)
+        {
+            var timeSpanExpression = timeSpan.Humanize(culture: culture, maxUnit: maxUnit, toWords: toWords);
+
+            var cultureFormatter = Configurator.GetFormatter(culture);
+            var ageExpression = string.Format(cultureFormatter.TimeSpanHumanize_Age(), timeSpanExpression);
+
+            return ageExpression;
+        }
+
         static IEnumerable<string> CreateTheTimePartsWithUpperAndLowerLimits(TimeSpan timespan, CultureInfo culture, TimeUnit maxUnit, TimeUnit minUnit, bool toWords = false)
         {
             var cultureFormatter = Configurator.GetFormatter(culture);
--- a/readme.md
+++ b/readme.md
@@ -444,7 +444,18 @@ TimeSpan.FromMilliseconds(1299630020).Humanize(3, collectionSeparator: null) => 
 If words are preferred to numbers, a `toWords: true` parameter can be set to convert the numbers in a humanized TimeSpan to words:
 ```C#
 TimeSpan.FromMilliseconds(1299630020).Humanize(3, toWords: true) => "two weeks, one day, one hour"
-````
+```
+
+By calling `ToAge`, a `TimeSpan` can also be expressed as an age.
+For cultures that do not define an age expression, the result will be the same as calling `Humanize` _(but with a default `maxUnit` equal to `TimeUnit.Year`)_. 
+
+```C#
+// in en-US culture
+TimeSpan.FromDays(750).ToAge() => "2 years old"
+
+// in fr culture
+TimeSpan.FromDays(750).ToAge() => "2 ans"
+```
 
 ### <a id="humanize-collections">Humanize Collections</a>
 You can call `Humanize` on any `IEnumerable` to get a nicely formatted string representing the objects in the collection. By default `ToString()` will be called on each item to get its representation but a formatting function may be passed to `Humanize` instead. Additionally, a default separator is provided ("and" in English), but a different separator may be passed into `Humanize`.
--- a/src/Humanizer.Tests.Shared/TimeSpanHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/TimeSpanHumanizeTests.cs
@@ -29,7 +29,7 @@
         [InlineData(365 + 365 + 365 + 365 + 1, "4 years")]
         [InlineData(365 + 365 + 365 + 365 + 366, "4 years, 11 months, 30 days")]
         [InlineData(365 + 365 + 365 + 365 + 366 + 1, "5 years")]
-        public void Year(int days, string expected)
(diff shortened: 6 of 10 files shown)


hazzik:
I think it would be better if the addition of `" old"` is added through the resource rather than logic checking for a specific culture. Because it is not strictly unique to English to add "old" modifier to describe the age. It seems applies to some Germanic languages as well:

- Afrikaans/Dutch: twee jaar **oud** 
- Danish/Norwegian: to år **gammel** 
- Swedish: två år **gammal** 
- German: zwei Jahre **alt**

src/Humanizer/TimeSpanHumanizeExtensions.cs
```diff
@@ -53,6 +53,28 @@ public static string Humanize(this TimeSpan timeSpan, int precision, bool countE
             return ConcatenateTimeSpanParts(timeParts, culture, collectionSeparator);
         }
 
+        public static string ToAge(this TimeSpan timeSpan, CultureInfo culture = null)
+        {
+            culture ??= CultureInfo.CurrentUICulture;
+
+            var ageExpression = timeSpan.Humanize(maxUnit: TimeUnit.Year);
+            if (culture.UsesEnglishLanguage())
+                ageExpression += " old";
```
hazzik, line 62: Please change to using resource string instead.

hazzik (changes requested):

src/Humanizer/TimeSpanHumanizeExtensions.cs
louis-z: That's a good idea. I can add something like the `TimeSpanHumanize_AgeSuffix` resource.

Because most languages do not have such a suffix, I'm thinking of making the default resources' TimeSpanHumanize_AgeSuffix = "", but since the default resources correspond to "en-US" it means I'll need to introduce Resources.en.resx and Resources.en-US.resx to override that resource, right? Is there an easier way?

hazzik: The thing is that it might be not a suffix but something else in some random language. So, I think it would be better to have a formatting string like `”{0} age”`.

Yes you would need to introduce default resource with `"{0}”` and appropriate resource for `en` culture.

louis-z: Ok so I introduced `TimeSpanHumanize_Age` & `TimeSpanHumanize_PreHyphenatedAge` format resources.

- Defined default values (in Resources.resx) as `"{0}"` and `""`, respectively
- Had to add Resources.en.resx and Resources.en-US.resx, to override the defaults
- Adapted nuspec files accordingly
- Simplified default (en-US) tests and moved tests for other English-based cultures to another file

src/Humanizer/Localisation/Formatters/IFormatter.cs
```diff
@@ -43,6 +43,23 @@ public interface IFormatter
...
+        /// <returns>Age format</returns>
+        string TimeSpanHumanize_Age();
+
+        /// <summary>
+        /// Returns the age format that converts a humanized TimeSpan string to an hyphenated age expression,
+        /// before spaces are actually replaced by hyphens.
+        /// </summary>
+        /// <returns>Pre-hyphenated age format</returns>
+        /// <remarks>
+        /// In all languages but English an empty string is returned, as the notion of hyphenated age does not exist.
```
louis-z, line 59: Hopefully I'm right about this. 😄

louis-z: Or perhaps I should (humbly) rephrase to `In most languages but English...`

src/Humanizer/TimeSpanHumanizeExtensions.cs
louis-z: After further consideration, I think I went about it the wrong way. 🤦‍♂️ 

Conceptually it made sense to have in `Resources.resx` the most common age formats, i.e.
- the basic age format being "{0}", i.e. the same as what `TimeSpan.Humanize()` returns
- the pre-hyphenated age format being "", because in most languages the hyphenated age makes no sense and it too should just be an empty string, right?

The problem is that `Resources.resx` happens to contain the en-US culture's resources. So with the above changes I had to introduce a `Resources.en-US.resx` file to override the default formats, but that's actually adding confusion and forcing people who just want the en-US resources to not only install Humanizer.Core, but also the eventual en-US NuGet package. Yikes...

I therefore made the following changes (which I haven't pushed yet):
- Moved to `Resources.resx` the en-US age formats
- Added to each of the other resx files the so-called most common age formats _(note that people with knowledge of other cultures may eventually need to revisit those resources, especially the Germanic ones mentioned previously)_

Let me know @hazzik whether I should push those changes.

hazzik:
Ok, I've read some info about the hyphenated age. So, hyphenated age is an age adjective. These do exist in languages other than English, in my native Russian for example:

- пятидесяти**летний** мужчина - fifty-years-old man.

"**летний**" here means "years-old" and "пятидесяти" - is an age component.

Similar for German: "fünfzig**jähr**iger"

So I think: we could accept the `ToAge()` extension as is and give a little more thought to `ToHyphenatedAge()` extension.

The points to consider:

- more appropriate name?
- how to support different languages?

src/Humanizer/TimeSpanHumanizeExtensions.cs
hazzik: Sorry @louis-z I don't really know what would be the best strategy to deal with resources.

src/Humanizer/Localisation/Formatters/IFormatter.cs
```diff
@@ -43,6 +43,23 @@ public interface IFormatter
         /// <returns></returns>
         string TimeSpanHumanize(TimeUnit timeUnit, int unit, bool toWords = false);
 
+        /// <summary>
+        /// Returns the age format that converts a humanized TimeSpan string to an age expression.
+        /// For instance, in English that format adds the " old" suffix, so that "40 years" becomes "40 years old".
+        /// </summary>
+        /// <returns>Age format</returns>
+        string TimeSpanHumanize_Age();
```
hazzik, line 51: This is a breaking change, so I'm not sure if this would be allowed. I think for the purpose of localization it would be better have it as a separate interface. @clairernovotny what do you think?

NuSpecs/Humanizer.Core.en-US.nuspec
```diff
@@ -0,0 +1,25 @@
...
+    <license type="expression">MIT</license>
+    <repository type="$RepositoryType$" url="$RepositoryUrl$" commit="$RepositoryCommit$" />
+    <language>en-US</language>
+    <dependencies>
+       <dependency id="Humanizer.Core" version="[$version$]" />
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Humanizer\bin\Release\netstandard1.0\en-US\*.*" target="lib\netstandard1.0\en-US" />
+    <file src="Humanizer\bin\Release\netstandard2.0\en-US\*.*" target="lib\netstandard2.0\en-US" />
```
clairernovotny, line 22: needs a net6.0 target now -- please rebase on main

NuSpecs/Humanizer.Core.en.nuspec
```diff
@@ -0,0 +1,25 @@
...
+    <license type="expression">MIT</license>
+    <repository type="$RepositoryType$" url="$RepositoryUrl$" commit="$RepositoryCommit$" />
+    <language>en</language>
+    <dependencies>
+       <dependency id="Humanizer.Core" version="[$version$]" />
+    </dependencies>
+  </metadata>
+  <files>
+    <file src="Humanizer\bin\Release\netstandard1.0\en\*.*" target="lib\netstandard1.0\en" />
+    <file src="Humanizer\bin\Release\netstandard2.0\en\*.*" target="lib\netstandard2.0\en" />
```
clairernovotny, line 22: Needs a net6 version

src/Humanizer/Localisation/Formatters/IFormatter.cs
clairernovotny: In this case, I think it's okay to add a new method to the interface--I don't expect many/any custom implementations. 

The primary concern behind breaking changes is that newer versions can be used with code compiled against the old one -- and adding a method to the interface is okay there.

clairernovotny:
Why do we need both a new `en` and `en-US` resource package? One of them will be included in the main dlll as embedded resources; which one is it?

NuSpecs/Humanizer.Core.en-US.nuspec
louis-z: I will gladly do it, if you confirm I've chosen the right approach right off the bat. See [here](https://github.com/Humanizr/Humanizer/pull/1068#discussion_r641011389). Otherwise I can push the changes I mentioned there.

louis-z:
> Why do we need both a new `en` and `en-US` resource package? One of them will be included in the main dlll as embedded resources; which one is it?

I must admit I wasn't too sure about that myself... But if I modify the PR according to what I proposed [here](https://github.com/Humanizr/Humanizer/pull/1068#discussion_r641011389), then this question will become irrelevant, right? 😄

NuSpecs/Humanizer.Core.en-US.nuspec
louis-z: Ok, just rebased on main.

NuSpecs/Humanizer.Core.en.nuspec
louis-z: Done

src/Humanizer/TimeSpanHumanizeExtensions.cs
louis-z: So I finally went ahead with the proposed changes, to simplify resource management:
- Made the English age and hyphenated age formats the default (i.e. through `Resources.resx`)
- Overrode those settings for French culture, as an example. For most other cultures, similar changes will be needed (should I submit them as part of this PR?)

src/Humanizer.Tests.Shared/Localisation/fr/TimeSpanHumanizeTests.cs
```diff
@@ -155,6 +155,30 @@ public void MillisecondsToWords(int ms, string expected)
...
 
+        [Theory]
+        [InlineData(4, false, "4 jours")]
+        [InlineData(23, false, "3 semaines")]
+        [InlineData(64, false, "2 mois")]
+        [InlineData(367, true, "un an")]
+        [InlineData(750, true, "deux ans")]
+        public void Age(int days, bool toWords, string expected)
+        {
+            var actual = TimeSpan.FromDays(days).ToAge(toWords: toWords);
```
aloisdg, line 166: "12 mois / 24 mois", pour l'âge on compte en mois le plus souvent au début.


For babies less than 2 years old, we count in months instead of year in France.

louis-z:
Hey @clairernovotny  & @hazzik  

I finally decided to drop the implementation of the so-called hyphenated (aka noun or pre-noun modifier) age expression, as it was too controversial. Can we now agree on this simpler PR? If not, please advise so I may make the necessary changes to have this PR accepted and merged.

louis-z:
@SimonCropp, do you think I should resurrect this PR, at least in its simplified form (refer to [this comment](https://github.com/Humanizr/Humanizer/pull/1068#issuecomment-1075726665))?

SimonCropp:
@louis-z  yeah i think a simpler implementation with a subset would be easier to get in 

should we close this one for now?

src/Humanizer.Tests.Shared/Localisation/fr/TimeSpanHumanizeTests.cs
hazzik: I think this is fine for start. Can be done as an improvement later.

src/Humanizer/Localisation/Resources.cs
```diff
@@ -21,5 +21,28 @@ public static string GetResource(string resourceKey, CultureInfo culture = null)
...
+        /// <param name="culture">The culture of the resource to retrieve. If not specified, current thread's UI culture is used.</param>
+        /// <param name="result">The value of the resource localized for the specified culture if found; null otherwise.</param>
+        /// <returns>true if the specified string resource was found for the given culture; otherwise, false.</returns>
+        /// <remarks>
+        /// (In .NET Standard 1.0, this method will not strictly adhere to the above contract.
+        /// It may fall back on the default culture.)
+        /// </remarks>
+        public static bool TryGetResource(string resourceKey, CultureInfo culture, out string result)
+        {
+#if NETSTANDARD1_0
```
hazzik, line 38: `NETSTANDARD1_0` has been dropped, so this is not needed anymore.

hazzik:
@louis-z could you please resolve conflicts and (re-)sign the cla?

hazzik (approved):

SimonCropp:
@louis-z can you add an explanation of the feature to the readme

louis-z:
> @louis-z can you add an explanation of the feature to the readme

Done

hazzik (approved):

PR #1105, opened 2021-07-03
Add German translations for ByteSize

hangy:
 - [X] Implementation is clean

For reference, see [Wikipedia](https://de.wikipedia.org/w/index.php?title=Byte&oldid=212871867#Vergleichstabelle).

### DIFF
--- a/src/Humanizer.Tests.Shared/Bytes/ToFullWordsTests.cs
+++ b/src/Humanizer.Tests.Shared/Bytes/ToFullWordsTests.cs
@@ -25,6 +25,7 @@ using Xunit;
 
 namespace Humanizer.Tests.Bytes
 {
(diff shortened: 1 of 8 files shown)


src/Humanizer.Tests.Shared/Bytes/ToFullWordsTests.cs
```diff
@@ -25,6 +25,7 @@
 
 namespace Humanizer.Tests.Bytes
 {
+    [UseCulture("en")]
```
hangy, line 28: I had to change this to use the `en` culture, because the tests otherwise don't run on a system running in a `de` culture setting.

clairernovotny:
Can you please rebase this?

PR #1106, opened 2021-07-03
Convert TimeUnit to symbols

hangy:
 - [X] Implementation is clean


hangy:
OK, this probably needs some discussion. 😊

I want to use localized versions of `TimeUnit` to localize `ByteRate.Humanize`, which currently only returns English time units. https://github.com/Humanizr/Humanizer/blob/d0737907c184cc1cd196e092ea42a53f033387cf/src/Humanizer/Bytes/ByteRate.cs#L51-L78

I assume that a general method to get a symbol from the `TimeUnit` could be useful, which is why I added the `TimeUnitToSymbolExtensions.ToSymbol` method. However, I'm not quite sure how/if it was okay to leak `Humanizer.Localisation.TimeUnit` to a broader namespace. I'm open for any suggestions. 😊

src/Humanizer/Properties/Resources.resx
```diff
@@ -711,4 +711,28 @@
...
+    <value>d</value>
+  </data>
+  <data name="TimeUnit_Week" xml:space="preserve">
+    <value>week</value>
+  </data>
+  <data name="TimeUnit_Month" xml:space="preserve">
+    <value>mo</value>
+  </data>
+  <data name="TimeUnit_Year" xml:space="preserve">
+    <value>a</value>
```
hangy, line 736: According to Wikipedia, `a` is a common year unit symbol in English, but `y` is apparently common, too. Maybe a native speaker could weigh in on this?

PR #1114, opened 2021-08-06
Fixed default number format

YohDeadfall:
This PR fixes an issue happening only when `ToString()` is invoked on a `ByteSize`. Since parameterless `ToString` is optimized and doesn't invoke `ToString(string format)` it uses `{0} {1}` to create an output string, but this format has no information on how a value should be displayed and prints a number with high precision.

### DIFF
--- a/src/Humanizer.Tests.Shared/Bytes/ToStringTests.cs
+++ b/src/Humanizer.Tests.Shared/Bytes/ToStringTests.cs
@@ -37,6 +37,7 @@ namespace Humanizer.Tests.Bytes
         [Fact]
         public void ReturnsDefaultNumberFormat()
         {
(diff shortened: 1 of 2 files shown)


src/Humanizer.Tests.Shared/Bytes/ByteRateTests.cs
```diff
@@ -47,7 +47,7 @@ public void TimeUnitTests(long megabytes, double measurementIntervalSeconds, Tim
         }
 
         [Theory]
-        [InlineData(19854651984, 1, TimeUnit.Second, null, "18.4910856038332 GB/s")]
+        [InlineData(19854651984, 1, TimeUnit.Second, null, "18.49 GB/s")]
```
YohDeadfall, line 50: Weird. This line isn't present in `main`.

PR #1116, opened 2021-08-17
Feature/icelandic language support

Kristinn-Stefansson:
As a native Icelandic speaker I wanted to contribute support for the language. Converting numeric data to words was my main aspiration so localisations of ToWords and ToOrdinalWords was the core aim, with Dates, TimeSpans, Headings and Bytes then being straightforward enough to add along the way. Icelandic has a few nuances there that some other North Germanic languages have left behind, even though heavier levels of inflection including four cases and noun declension are not in the picture. Still, I hope the submitted implementation is both transparent and in line with the established code. I have run a copy of the build process to validate that all tests pass on the PR, and did some benchmarks on the side to try and make sure the performance compared favorably with a selection of other language implementations.

The PR checklist: 
 - [X] Implementation is clean
 - [ ] Link to the issue(s) you're fixing from your PR description. Use `fixes #<the issue number>`


src/Humanizer/Properties/Resources.is.resx
```diff
@@ -0,0 +1,540 @@
...
+    <value>fyrir {0} sekúndu</value>
+  </data>
+    <data name="DateHumanize_MultipleSecondsFromNow_Plural" xml:space="preserve">
+    <value>eftir {0} sekúndur</value>
+  </data>
+  <data name="DateHumanize_MultipleSecondsFromNow_Singular" xml:space="preserve">
+    <value>eftir {0} sekúndu</value>
+  </data>
+    <data name="DateHumanize_MultipleYearsAgo_Above20" xml:space="preserve">
+    <value>{0} years ago</value>
```
clairernovotny, line 322: Is this correct?

src/Humanizer/Properties/Resources.is.resx
```diff
@@ -0,0 +1,540 @@
...
+    <value>fyrir {0} mánuði</value>
+  </data>
+      <data name="DateHumanize_MultipleMonthsFromNow_Plural" xml:space="preserve">
+    <value>eftir {0} mánuði</value>
+  </data>
+  <data name="DateHumanize_MultipleMonthsFromNow_Singular" xml:space="preserve">
+    <value>eftir {0} mánuð</value>
+  </data>
+    <data name="DateHumanize_MultipleSecondsAgo_Above20" xml:space="preserve">
+    <value>{0} seconds ago</value>
```
clairernovotny, line 307: Is this correct?

src/Humanizer/Properties/Resources.is.resx
```diff
@@ -0,0 +1,540 @@
...
+    <value>{0} mínútur</value>
+  </data>
+  <data name="TimeSpanHumanize_MultipleSeconds_Plural" xml:space="preserve">
+    <value>{0} sekúndur</value>
+  </data>
+  <data name="TimeSpanHumanize_MultipleMilliseconds_Plural" xml:space="preserve">
+    <value>{0} millisekúndur</value>
+  </data>
+  <data name="DateHumanize_MultipleDaysAgo_Above20" xml:space="preserve">
+    <value>{0} days ago</value>
```
clairernovotny, line 253: Is this correct?

clairernovotny (commented):
This looks great, thank you! I noticed a few things in the ResX that did not appear to be translated; are those accurate?

src/Humanizer/Properties/Resources.is.resx
Kristinn-Stefansson: Thanks @clairernovotny, good catch. I originally wanted to trace their use and see if they would be caught in tests but then forgot to finish cleaning up. I have translated them, and pushed a new commit.

src/Humanizer/Properties/Resources.is.resx
Kristinn-Stefansson: See previous revolved comment.

clairernovotny:
The one remaining piece is to add a reference to this language in the main metapackage, like this:
https://github.com/Humanizr/Humanizer/blob/main/NuSpecs/Humanizer.nuspec#L36

Once that's done, I can merge this and get a release out.

PR #1117, opened 2021-08-17
feat: Improve FromAbbreviatedHeading for localized headings

hangy:
 - [X] Implementation is clean

### DIFF
--- a/src/Humanizer/HeadingExtensions.cs
+++ b/src/Humanizer/HeadingExtensions.cs
@@ -70,17 +70,28 @@ namespace Humanizer
         /// <summary>
         /// Returns a heading based on the short textual representation of the heading.
         /// </summary>
+        /// <param name="culture">The culture of the heading</param>
         /// <returns>The heading. -1 if the heading could not be parsed.</returns>
-        public static double FromAbbreviatedHeading(this string heading)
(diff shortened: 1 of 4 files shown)


src/Humanizer/HeadingExtensions.cs
```diff
@@ -70,17 +71,61 @@ public static char ToHeadingArrow(this double heading)
         /// <summary>
         /// Returns a heading based on the short textual representation of the heading.
         /// </summary>
+        /// <param name="culture">The culture of the heading</param>
         /// <returns>The heading. -1 if the heading could not be parsed.</returns>
-        public static double FromAbbreviatedHeading(this string heading)
+        public static double FromAbbreviatedHeading(this string heading, CultureInfo culture = null)
```
clairernovotny, line 76: This should be an additional method/override, so we don't break existing people.

clairernovotny (commented):
LGTM with the one fix

src/Humanizer/HeadingExtensions.cs
hangy: Thanks! I think it should be fixed in 9d48857d9bc78cafdde18efcc0d8093674e8c6fd.

Issue #1081, opened 2021-06-07
Add support for `TimeOnly` to display clock times "Three O'Clock", etc


clairernovotny:
#1080 Adds partial support for `TimeOnly`, to display it in terms of differences, we should have a way to turn a `TimeOnly` value into something more natural, like clock time.

neilboyd:
This is fixed in #1134

clairernovotny:
Thank you for proposing natural clock-time support. This was implemented by #1134 and shipped in Humanizer 2.13.14; current releases expose `TimeOnly.ToClockNotation()`.

PR #1134, opened 2021-10-27
Add support to humanizing a TimeOnly as a readable clock notation

strobelt:
Add support to humanizing a TimeOnly as a readable clock notation string as asked on #1081 
Uses the Converter architecture based on the DateOnly ToOrdinal converter and extension


strobelt:
Great! I made the change and went with the enum to make it more clear when using the converter.
What do you think?

src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.approve_public_api.approved.txt
```diff
@@ -1945,4 +1950,16 @@ namespace Humanizer.Localisation.Ordinalizers
         string Convert(int number, string numberString);
         string Convert(int number, string numberString, Humanizer.GrammaticalGender gender);
     }
+}
+namespace Humanizer.Localisation.TimeToClockNotation
+{
+    public enum ClockNotationRounding
```
clairernovotny, line 1956: This enum should be in the root `Humanizer` namespace so that people don't need a deep include for this

clairernovotny (changes requested):

clairernovotny:
Ping @strobelt? I'd love to get a new release out that includes these changes!

strobelt:
Hi @clairernovotny! I just updated the PR. Thanks for the heads-up with this

clairernovotny (approved):

PR #1142, opened 2021-10-29
Fixed merge issue with lost test case change

YohDeadfall:
Don't know how, but 7f86099b42ab8535f08b99256498aded788965d0 was lost when merging #1114. This PR changes the test case once more.

cc @clairernovotny


clairernovotny:
Thanks!

PR #1164, opened 2022-01-10
fixes #1161

### LINKED ISSUE #1161: Number to word and Indian numeral system
How to enable Indian numeral system in **_Number to Word_** function 

Ref:
https://en.wikipedia.org/wiki/Indian_numbering_system
https://www.cuemath.com/numbers/indian-numeral-system/

biju-ps:
 - [x] Implementation is clean

### DIFF
--- a/src/Humanizer.Tests.Shared/Localisation/en-IN/NumberToWordsTests.cs
+++ b/src/Humanizer.Tests.Shared/Localisation/en-IN/NumberToWordsTests.cs
@@ -43,6 +43,8 @@ namespace Humanizer.Tests.Localisation.enIN
         [InlineData(12345678, "one crore twenty three lakh forty five thousand six hundred and seventy eight")]
         [InlineData(123456789, "twelve crore thirty four lakh fifty six thousand seven hundred and eighty nine")]
         [InlineData(1234567890, "one hundred and twenty three crore forty five lakh sixty seven thousand eight hundred and ninety")]
+        [InlineData(1000000000000, "one lakh crore")]
+        [InlineData(45678912345678, "forty five lakh sixty seven thousand eight hundred and ninety one crore twenty three lakh forty five thousand six hundred and seventy eight")]
         [InlineData(-7516, "(Negative) seven thousand five hundred and sixteen")]
 
         public void ToWords(long number, string expected)
--- a/src/Humanizer.Tests.Shared/Localisation/en-IN/NumberToWordsTests.cs
+++ b/src/Humanizer.Tests.Shared/Localisation/en-IN/NumberToWordsTests.cs
@@ -1,0 +1,62 @@
+﻿using Xunit;
+
+namespace Humanizer.Tests.Localisation.enIN
+{
+    [UseCulture("en-IN")]
+    public class NumberToWordsTests
+    {
+        [Theory]
+        [InlineData(0, "")]
+        [InlineData(1, "one")]
+        [InlineData(10, "ten")]
+        [InlineData(11, "eleven")]
+        [InlineData(20, "twenty")]
+        [InlineData(122, "one hundred and twenty two")]
+        [InlineData(3501, "three thousand five hundred and one")]
+        [InlineData(100, "one hundred")]
+        [InlineData(1000, "one thousand")]
+        [InlineData(1001, "one thousand one")]
+        [InlineData(100000, "one lakh")]
+        [InlineData(1000000, "ten lakh")]
+        [InlineData(10000000, "one crore")]
+        [InlineData(100000000, "ten crore")]
+        [InlineData(1000000000, "one hundred crore")]
+        [InlineData(111, "one hundred and eleven")]
+        [InlineData(1111, "one thousand one hundred and eleven")]
+        [InlineData(111111, "one lakh eleven thousand one hundred and eleven")]
+        [InlineData(1111111, "eleven lakh eleven thousand one hundred and eleven")]
+        [InlineData(11111111, "one crore eleven lakh eleven thousand one hundred and eleven")]
+        [InlineData(111111111, "eleven crore eleven lakh eleven thousand one hundred and eleven")]
+        [InlineData(1111111111, "one hundred and eleven crore eleven lakh eleven thousand one hundred and eleven")]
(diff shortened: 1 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/Humanizr/Humanizer?pullRequest=1164) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/Humanizr/Humanizer?pullRequest=1164) before we can accept your contribution.<br/><br/>:x: biju-ps [sign now](https://cla.dotnetfoundation.org/Humanizr/Humanizer?pullRequest=1164)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/Humanizr/Humanizer?pullRequest=1164) it.</sub>

src/Humanizer.Tests.Shared/Localisation/en-IN/NumberToWordsTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        [InlineData(122, "one hundred and twenty two")]
+        [InlineData(3501, "three thousand five hundred and one")]
+        [InlineData(100, "one hundred")]
+        [InlineData(1000, "one thousand")]
+        [InlineData(1001, "one thousand one")]
+        [InlineData(100000, "one lakh")]
+        [InlineData(1000000, "ten lakh")]
+        [InlineData(10000000, "one crore")]
+        [InlineData(100000000, "ten crore")]
+        [InlineData(1000000000, "one hundred crore")]
```
vatsan-madhavan, line 23: The rules IIRC are:
- Upto 99,999 - identical to US place value system.
- Beyond 1E5, step-size is 1E2 (and not 1E3 like US). 
- 1,00,000 = 1 `lakh`
- 1,00,00,000 = 1 `crore` = 100 `lakh`
- 1,00,00,00,000 = 1 `arab` = 100 `crore`
- 1,00,00,00,00,000 = 1 `kharab` = 100 `arab`
- There are more of these - far less used ones (and less well-known), like `nil` (100 `kharab`, pronounced like *neel*), `padma` (*100 `nil`*) etc. 

The key challenge when building software is deciding how to deal with larger values like `nil`, `padma` etc. They're rarely used, and in practice native speakers simply express `1000 * 1 kharab` as `thousand kharab` (instead of `10 nil`). 

Choosing between pedantic accuracy vs. comprehension is a design-choice software has to make when dealing with numbers > *99 `kharab`*. 

In practice, the usage frequency (as well as familiarity) of `arab`, `kharab` (and larger ones) drops - `crore` is the most commonly used/known one. An argument could be made that usable design would stop at `crore` and ignore `arab` and larger names. That said, `arab`, `kharab` are in extant use (both my experience/memory; I also checked with couple of native Hindi speakers to confirm). IMHO think the right balance would be to retain until `kharab`. 

==
Assuming the choice is to incorporate `arab` and `kharab`, I believe `1000000000` should be `one arab` (rather than `one hundred crore`).

clairernovotny: Thanks for the analysis @vatsan-madhavan! 

@biju-ps Could you please update and we can get this merged?

src/Humanizer/Localisation/NumberToWords/IndianNumberToWordsConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+            if (number == 0)
+                return "";
+            if (number < 0)
+                return "(Negative) " + NumberToText(-number);
+            if (number == 0)
+                return "";
+            if (number <= 19)
+                return new[]
+                {
+                "one", "two", "three", "four", "five", "six", "seven", "eight",
```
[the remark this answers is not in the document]
biju-ps: done

src/Humanizer/Localisation/NumberToWords/IndianNumberToWordsConverter.cs
```diff
@@ -0,0 +1,76 @@
...
+                "nine", "ten", "eleven", "twelve", "thirteen", "fourteen", "fifteen", "sixteen",
+                "seventeen", "eighteen", "nineteen"
+            }[number - 1] + " ";
+            if (number <= 99)
+
+                return new[]
```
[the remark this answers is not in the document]
biju-ps: done

src/Humanizer.Tests.Shared/Localisation/en-IN/NumberToWordsTests.cs
vatsan-madhavan: Thanks for following up on this. What a bummer about COVID-19 - glad to see that you're better now!

I agree that this phenomenon (of expressing everything in terms of `lakh`, and `crore`, and not leveraging `arab`, `kharab`) is not new - this is in line with my experience, and also a couple of folks I spoke about this (to confirm my assumptions, before I posted my earlier comment) thought so as well.

It's not very different than saying that many (but not all) are familiar with `quadrillion` (which comes after `trillion`), but few rarely use it. `quintillion` etc. are far less known, and very rarely used. That said, the question becomes something like this - in a programming library, should one emit `ten thousand trillion`, or should the library emit `10 quadrillion`. This is ultimately a value judgment and depends on your design principles and use-goals.

If you and @clairernovotny decide to represent everything in terms of `crore`, I don't think it would be *wrong* per-se - just a different design choice.

Perhaps what you need here is the ability to switch between *strict* and *common speech* modes, each with different outputs.

clairernovotny:
When it comes to localization, I don't have an opinion as it's not my tounge. I'll defer to those who know the language and culture better and go with that.

src/Humanizer.Tests.Shared/Localisation/en-IN/NumberToWordsTests.cs
clairernovotny: @vatsan-madhavan thanks for the explanation. I think we can merge this as-is and create a follow-up issue to support the two different modes if someone wants to tackle it.

PR #1176, opened 2022-02-08
fix warnings

neilboyd:
Fix build warnings so that the CI build is clean.

---

 - [x] Implementation is clean
     - see #1177

### DIFF
--- a/src/Humanizer/ClockNotationRounding.cs
+++ b/src/Humanizer/ClockNotationRounding.cs
@@ -1,8 +1,18 @@
 ﻿namespace Humanizer
 {
+    /// <summary>
+    /// Options for specifying readable clock notation
+    /// </summary>
     public enum ClockNotationRounding
     {
+        /// <summary>
+        /// Do not round minutes
+        /// </summary>
         None,
+
+        /// <summary>
+        /// Round time to nearest five minutes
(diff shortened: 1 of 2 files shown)


src/Humanizer/Localisation/ResourceKeys.TimeUnitSymbol.cs
```diff
@@ -16,8 +16,6 @@ public static class TimeUnitSymbol
             /// Generates Resource Keys according to convention.
             /// </summary>
             /// <param name="unit">Time unit, <see cref="TimeUnit"/>.</param>
-            /// <param name="count">Number of units, default is One.</param>
-            /// <param name="toWords">Result to words, default is false.</param>
```
neilboyd, line 20: This change was introduced in #1106

src/Humanizer/ClockNotationRounding.cs
```diff
@@ -1,8 +1,18 @@
 ﻿namespace Humanizer
 {
+    /// <summary>
+    /// Options for specifying readable clock notation
+    /// </summary>
     public enum ClockNotationRounding
```
neilboyd, line 6: This change was introduced in #1134

PR #1178, opened 2022-02-17
special case for letter 's'

### LINKED ISSUE #1154: [BUG] Index out of range during `Singularize`
When I execute
```csharp
var result = "S".Singularize();
```
it throws this exception:
```
Exception: Index was outside the bounds of the array.
   at System.String.get_Chars(Int32 index)
   at Humanizer.Inflections.Vocabulary.MatchUpperCase(String word, String replacement) in /_/src/Humanizer/Inflections/Vocabulary.cs:line 158
   at Humanizer.Inflections.Vocabulary.Singularize(String word, Boolean inputIsKnownToBePlural, Boolean skipSimpleWords) in /_/src/Humanizer/Inflections/Vocabulary.cs:line 104
   at Humanizer.InflectorExtensions.Singularize(String word, Boolean inputIsKnownToBePlural, Boolean skipSimpleWords) in /_/src/Humanizer/InflectorExtensions.cs:line 54
```

neilboyd:
Handle singularizing and pluralizing of letter 's'.

It is not sufficient to just handle singularizing of single letter words because letter 's' has a special meaning when singularizing and pluralizing.

Fixes #1154

---

 - [x] Implementation is clean

### DIFF
--- a/src/Humanizer.Tests.Shared/InflectorTests.cs
+++ b/src/Humanizer.Tests.Shared/InflectorTests.cs
@@ -71,6 +71,19 @@ namespace Humanizer.Tests
             Assert.Equal(singular, plural.Singularize(skipSimpleWords: true));
         }
 
+        [Theory]
+        [InlineData("a")]
+        [InlineData("A")]
+        [InlineData("s")]
+        [InlineData("S")]
+        [InlineData("z")]
+        [InlineData("Z")]
+        [InlineData("1")]
+        public void SingularizeSingleLetter(string input)
+        {
+            Assert.Equal(input, input.Singularize());
+        }
+
         //Uppercases individual words and removes some characters 
         [Theory]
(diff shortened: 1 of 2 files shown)


src/Humanizer.Tests.Shared/InflectorTests.cs
```diff
@@ -280,6 +280,7 @@ public IEnumerator<object[]> GetEnumerator()
             yield return new object[] { "alumna", "alumnae" };
             yield return new object[] { "alumnus", "alumni" };
             yield return new object[] { "fungus", "fungi" };
+
```
neilboyd, line 283: Break here because this is where the [original file](https://github.com/srkirkland/Inflector/blob/master/Inflector.Tests/PluralizeTests.cs) stopped.

PR #1183, opened 2022-02-23
time span format verbosity

### LINKED ISSUE #571: Minimize verbosity of Timespan formatting.
Is there a way to use duration abbreviations when Humanizing a TimeSpan? I'd like to use mins instead of minutes etc..

### LINKED ISSUE #642: Time units
Is it possible to print internal standard time units (http://www.ewh.ieee.org/soc/ias/pub-dept/abbreviation.pdf) instead of long units ('s' instead of 'second', 'h' instead of 'hour')?

### LINKED ISSUE #969: Abbreviate time units
I've read the the docs but did not find an option to abbreviate time units? 

E.g. millisecond ->ms, seconds -> sec, and so on

neilboyd:
Fixes #571
Fixes #642
Fixes #969

Based on [`HeadingStyle`](https://github.com/Humanizr/Humanizer/blob/main/src/Humanizer/HeadingExtensions.cs) (although that's inconsistent with [`WordForm`](https://github.com/Humanizr/Humanizer/blob/main/src/Humanizer/WordForm.cs)).

Replace `timeSpan.Humanize(toWords: true)` with `timeSpan.Humanize(timeSpanStyle: TimeSpanStyle.Words)`.
**This is a breaking change.**

`timeSpan.Humanize()` remains unchanged, although the optional parameter has changed from `toWords: false` to `timeSpanStyle: TimeSpanStyle.Full`.

Added `timeSpan.Humanize(timeSpanStyle: TimeSpanStyle.Abbreviated)` and `timeSpan.Humanize(timeSpanStyle: TimeSpanStyle.Short)`.
These return English regardless of the locale.

**Question**: how do you normally handle translation in this project? I could add tests for every language, where `Abbreviated` and `Short` return English for every locale, or not, or I could create translations (and tests) for every language (I could make a best guess for Latin and Cyrillic alphabets, but not the others).

**Question**: should we use [uppercase](https://www.w3.org/TR/NOTE-datetime) for `Y`, `D`, `W`? (`M` is already uppercase in order to disambiguate with minute)

---

 - [x] Implementation is clean


neilboyd:
I think it's ready, but there are some questions in the description

SimonCropp:
@hazzik thoughts on this one?

one note is that it significantly increase the number of entries in all the resource files. but if the entry is missing it can sensibly fall back to an acceptable value

hazzik:
I'm refactoring this whole thing into using fewer resources:

`DateHumanize_MultipleDaysFromNow` (`"{0} days from now"`) would become:

`DateHumanize_Future` => `"{0} from now"` (new resource) and
`TimeSpanHumanize_MultipleDays` => `"{0} days"` (existing resource)

SimonCropp:
@hazzik awesome. love your work

hazzik:
I've tried to refactor *Ago and *FromNow. Unfortunately it is not an easy task :( Too many language specific rules.

clairernovotny:
Thank you for the extensive work on compact TimeSpan formatting. The resource-key/resx localization layer and many of the paths changed here were replaced by the YAML/source-generator architecture in #1688, so this implementation no longer applies. The abbreviated-duration idea remains valuable and should be brought forward as a focused proposal against the current architecture.

PR #1236, opened 2022-11-02
fixed Issue 1186

### LINKED ISSUE #1186: `DateOnly` has current date return 'today'
When you want to humanize the `DateOnly` that has current date return 'now'.it's better to return 'today'. 'now' Word referee current date and time but 'today' referee only current date

hemalshah28:
 - [x] Implementation is clean
 - [x] Link to the issue(s) you're fixing from your PR description. Use `fixes #1186 `

### DIFF
--- a/src/Humanizer.Tests.Shared/DateHumanizeDefaultStrategyTests.cs
+++ b/src/Humanizer.Tests.Shared/DateHumanizeDefaultStrategyTests.cs
@@ -100,7 +100,7 @@ namespace Humanizer.Tests
         [InlineData(1, "tomorrow")]
         [InlineData(10, "10 days from now")]
         [InlineData(27, "27 days from now")]
-        [InlineData(32, "one month from today")]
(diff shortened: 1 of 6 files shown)


behroozbc:
Hi, I think your implementation isn't accurate. You need to implement `DateOnly` and not use `DateTime`. Nevertheless, I see you use the `DateTime`.

behroozbc (changes requested):
Hi, I think your implementation isn't accurate. You need to implement `DateOnly` and not use `DateTime`. Nevertheless, I see you use the `DateTime`.

hemalshah28:
> Hi, I think your implementation isn't accurate. You need to implement `DateOnly` and not use `DateTime`. Nevertheless, I see you use the `DateTime`.
Ok, Let me look into it.

hemalshah28:
@behroozbc , please take a look and let me know. sorry for misunderstanding.

hemalshah28:
I was looking for some help.

PR #1299, opened 2023-10-09
From kebabed to pascalized fix Issue#1282

dramlian:
Adjusted regex to handle converting from kebabed to pascalized.
Pascalize now works properly when first character is the separator.
Pascalize also works if there are multiple character separators such as "--"/"__" or two spaces.
All additions are covered in unit tests, all tests have passed succesfully.


dramlian:
@dotnet-policy-service agree

SimonCropp:
thanks for the contribution

PR #1364, opened 2024-02-16
remove redundant alloc due to ToCharArray

SimonCropp:


src/Humanizer/StringHumanizeExtensions.cs
```diff
@@ -29,12 +29,16 @@ private static string FromPascalCase(string input)
...
+                {
+                    var value = match.Value;
+                    return value.All(char.IsUpper) &&
+                           (value.Length > 1 || (match.Index > 0 && input[match.Index - 1] == ' ') || value == "I")
+                        ? value
+                        : value.ToLower();
+                }));
 
-            if (result.Replace(" ", "").ToCharArray().All(c => char.IsUpper(c)) &&
+            if (result.Replace(" ", "").All(char.IsUpper) &&
```
hazzik, line 41: Depending on compiler the method group can be slower than a lambda due to boxing to a delegate.

SimonCropp: does that effect us when we always compile our dll with the latests c# and sdk?

hazzik: I don’t know. Probably not. It’s just a habit from old days.

hazzik (approved):

PR #1407, opened 2024-02-17
remove GitVersioning

SimonCropp:
i propose we move off GitVersion. for context i am one of the original creators of GitVersion. and i no longer use it in any of my projects, nor recommend it to others. GitVersion predates the the simplified versioning of the new sdk style projects. ie now we can have a single version number that is applied to nugets, assembly versions and metadata. and when GitVersion fails, due to some git tree structure it doesnt understand, it is very difficult to debug.

GitVersion  is also cognitive load for contributors/maintainers since it is a different approach from the standard "version in msbuild file" approach


clairernovotny:
Hey, please reverse this---lets chat if you have questions.

SimonCropp:
@clairernovotny would love to chat. ping me at simon.cropp@gmail.com to set up a call

PR #1413, opened 2024-02-18
Fix ByteSize implementation in regards to IEC specification

CollinAlpert:
This PR is targeted for the v3 release, since it contains a breaking change in how the `ByteSize` calculates sizes.

Until now, one kilobyte is considered to be 1024 bytes, which is incorrect. The code `Console.WriteLine(ByteSize.FromMegabytes(1).Bytes);` prints `1048576` instead of `1000000` which was initially very surprising to me. I have had countless discussions over this surprising behavior with my teammates.

This PR fixes the calculation and adds methods to aid in calculating binary instead of decimal powers.


CollinAlpert:
I will need for somebody to help me out with the `PublicApiApprovalTest`, since I have no idea what is going wrong there.

clairernovotny:
Thank you, @CollinAlpert, for the implementation and careful standards discussion here.

We carried the compatibility-safe part forward in #1851: explicit KiB/MiB/GiB/TiB APIs, parsing, formatting, and localization, while preserving Humanizer's existing KB-TB scaling. The global KB/MB semantic change in this branch would alter established shipped-v3 public behavior, so it is not being merged.

Your contribution is credited prominently in the replacement PR and your authorship is preserved on its implementation commit. Closing this PR in favor of #1851.

PR #1437, opened 2024-02-22
Enable Nullable ref types

SimonCropp:


hazzik:
I think you need to approve the API difference. 

I'm just wondering how did you determine where the types be nullable and where not?

SimonCropp:
fixed the api

> I'm just wondering how did you determine where the types be nullable and where not?

There are a few techniques

 * good tests can inform it. since where nulls are passed those parameters must be nullable. where argumentnull exceptions are asserted those parameters are not null
 * Optional parameters (where `parameter =null`) help a lot
 * any method that returns a null. then the return value is nullable
 * an understanding of the expectations public api can inform u where return values should not be null 

Then usually that is enough for the compiler to help you flow those nullables through the rest of the code

PR #1445, opened 2024-02-23
Updated Russian localization resources. fixes #1444

### LINKED ISSUE #1444: TimeSpan Humanize ignored toWords: false parameter
When executing the code:
```csharp
var oneDay = TimeSpan.FromDays(1);

oneDay.Humanize(
    culture: CultureInfo.GetCultureInfo("ru"),
    toWords: false
);
```
Result:
"один день"
Instead of the expected result:
"1 день"

This result is not only in Russian localization, but also in many others!
I think that you still need to take into account the toWords parameter with a single value, most likely you can fix it here:
https://github.com/Humanizr/Humanizer/blob/c915898b9ebf5b0ddde2384b284f02896cd29e11/src/Humanizer/Localisation/ResourceKeys.TimeSpanHumanize.cs#L27

P.S. Congratulations to everyone on the fact that the library has begun to be updated again 🥰

BloodShadow174:
Fixes #1444

### DIFF
--- a/src/Humanizer.Tests.Shared/Localisation/ru-RU/TimeSpanHumanizeTests.cs
+++ b/src/Humanizer.Tests.Shared/Localisation/ru-RU/TimeSpanHumanizeTests.cs
@@ -5,44 +5,49 @@
     {
         [Theory]
         [Trait("Translation", "Native speaker")]
-        [InlineData(366, "один год")]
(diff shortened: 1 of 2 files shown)


BloodShadow174:
@dotnet-policy-service agree

hazzik:
Thanks

PR #1454, opened 2024-02-24
use range indexes

hazzik:


SimonCropp:
i have been holding off doing this. since searching for SubString is an easy way to find places where span should be used

PR #1480, opened 2024-03-06
re-add gitversion

SimonCropp:


version.json
```diff
@@ -0,0 +1,10 @@
+{
+  "version": "2.14",
```
hazzik, line 2: I think we would need to bump the version as we did not follow the conventions.

hazzik (approved):

version.json
SimonCropp: lol. as always, thanks for checking my work :)

hazzik (approved):

PR #1482, opened 2024-03-08
drop net462 and net472

SimonCropp:


SimonCropp:
@clairernovotny as requested

PR #1504, opened 2024-05-28
Remove allocation from ByteSize.TryParse

stephentoub:


src/Humanizer/Bytes/ByteSize.cs
```diff
@@ -386,73 +387,82 @@ public int CompareTo(object? obj)
     public static bool TryParse(string? s, out ByteSize result) =>
         TryParse(s, null, out result);
 
-    public static bool TryParse(string? s, IFormatProvider? formatProvider, out ByteSize result)
+    public static bool TryParse(string? s, IFormatProvider? formatProvider, out ByteSize result) =>
+        TryParse(s.AsSpan(), formatProvider, out result);
+
+    static bool TryParse(CharSpan span, IFormatProvider? formatProvider, out ByteSize result)
```
hazzik, line 393: Wouldn't we benefit if this become public?

stephentoub: From my perspective, yes, which is why I factored it out into its own method. But I didn't want to presume anything about adding public surface area.

SimonCropp: yeah. i agree, it should be public

stephentoub: Added the public API

PR #1505, opened 2024-05-28
Replace relevant Substring use with AsSpan

stephentoub:


SimonCropp:
@stephentoub thanks :)

PR #1506, opened 2024-05-28
Cache Truncator singletons returned from static properties

stephentoub:


SimonCropp:
@stephentoub thanks :)

PR #1509, opened 2024-05-28
use InternalsVisibleTo MSBuild item

WeihanLi:
Simplify the `InternalsVisibleTo` usage with new features

- Use `InternalsVisibleTo` to generate the attribute, https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#internalsvisibleto
- Add the `PublicKey` MSBuild property to simplify the PublicKey configuration, https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#internalsvisibleto


clairernovotny:
Adding a strong name is a breaking change, which we don't want to do

PR #1512, opened 2024-05-30
Fix assert in GetRomanNumeralCharValue

stephentoub:
Should have been upper instead of lower. Not sure why CI didn't fail, unless it's only running tests against release?


SimonCropp:
>  running tests against release?

yep the ci build is built and tested in release mode

Issue #1529, opened 2024-07-24
Adding welsh (cy) support


IeuanWalker:
Hi, I've created a POC project for my workplace, and that piece of work is going ahead now. Our current app supports the Welsh language so looking to see if/ how I create a PR for it.

Currently, we only need support for `DateTimeOffset.Humanize` - 
![image](https://github.com/user-attachments/assets/9be57bbd-2035-41a1-bdd8-9d22e8d9c045)

So I'd like to do a PR to add support for that function, but would like to add support for all the features.

_Some background, I work for Cardiff city, which is the capital of Wales and the Welsh language. We have our own in-house translation team for Welsh language_

hangy:
I believe every language support is highly welcome! 👍🏻 Generally, it's not a problem to start with one feature for a language and to add others gradually. It looks like `DateTimeOffset.Humanize` is mostly based off of resource files, but unit tests to ensure that everything works as expected would be necessary. #924 looks like a good sample of adding support for a new culture.

IeuanWalker:
thanks @hangy, i've created a PR, would you know the answers to the questions i left in a comment on there?

clairernovotny:
Thank you for proposing Welsh support and for the earlier contribution. Welsh locale parity has now been implemented by #1770 on main. It has not yet been included in a numbered release.

PR #1533, opened 2024-08-16
Added welsh (cy)

IeuanWalker:
#1529

 - [x] Implementation is clean
 - [ ] Your PR is (re)based on top of the latest commits from the `main` branch (more info below)


IeuanWalker:
Hi, 2 things I'd like to do before changing this draft PR into a complete PR -

- Is there a way to get an output of all the translations being used by the library?
     - I see some languages have custom code to handle their unique differences - https://github.com/Humanizr/Humanizer/tree/main/src/Humanizer/Localisation
     - So I'd like to send an output of what this library would generate to our translation team and they can verify if everything looks ok as is, or if we'd need to look at doing something similar to what other languages have done
- Is there a standard set of tests you expect?
      - Or do you only expect tests if languages have implemented their own code?

hangy:
>     * Is there a way to get an output of all the translations being used by the library?
>       
>       * I see some languages have custom code to handle their unique differences - https://github.com/Humanizr/Humanizer/tree/main/src/Humanizer/Localisation
>       * So I'd like to send an output of what this library would generate to our translation team and they can verify if everything looks ok as is, or if we'd need to look at doing something similar to what other languages have done

Not that I know of. Creating all permutations for some languages with support for NumberOfWords, ByteSize, and so on sounds like a lot of information. I wonder if it might be possible to generate that with reflection, though.

>     * Is there a standard set of tests you expect?
>       
>       * Or do you only expect tests if languages have implemented their own code?

As far as I've seen, it's good practice to add some tests with new translations. You're correct that it might seem unnecessary if there's no custom code, but since the strings are also run through `string.Format` or are potentially concatenated in another way, having a test ensures that a change in the base code doesn't break the specific translation.

Do note that I'm just a random contributor and don't speak in any official capacity for the project. :)

IeuanWalker:
@clairernovotny  @SimonCropp  any chance this can be reviewed?

Currently using a custom nuget in a production application and have a few other projects i want to use this in but don't really want to keep copying the nuget to each solution.

clairernovotny:
Thank you for the Welsh localization contribution. PR #1770 has since added full Welsh locale parity using the current YAML/source-generator architecture, so this older implementation has been superseded.

PR #1537, opened 2024-08-24
Change TimeOnly.ToClockNotation to accept culture

RenNagasaki:
There are cases where your UI culture is not the culture you want to get the clock notation for. Added the functionality for that.
Existing implementations should not be influenced.

For me the problematic part is: I use the DefaultTimeOnlyToClockNotationConverter for english clock notation but my UI culture is 'de-DE'. I get for 08:15 -> 'a quarter past acht' instead of 'a quarter past eight'

 - [x] Implementation is clean
 - [] Xml documentation is added/updated for the addition/change


src/Benchmarks/TimeOnlyToClockNotationConverterBenchmarks.cs
```diff
@@ -1,4 +1,4 @@
-﻿[MemoryDiagnoser(false)]
+[MemoryDiagnoser(false)]
```
hangy, line 1: The first line of every file you modified seems to be changed. Is this a modified BOM? I'm not sure if there's a preferred version in the project, but you might want to check this.

RenNagasaki: All I know is I didn't actually touch that part. 🤔 
Has VS some kind of default setting? If yes then that's what changed.
I don't see any problems happening here though and kinda don't know how I could change that. Git inside VS doesn't show them as changed. 
grafik

clairernovotny:
Thank you for working on an explicit culture override for TimeOnly.ToClockNotation. The converter files and registration structure changed substantially in #1688 to YAML/source-generated phrase-clock engines, so this implementation no longer applies. The explicit-culture API idea is still valuable and should be reconsidered as a fresh change against the current architecture.

clairernovotny:
Thank you again, @RenNagasaki, for the explicit-culture design.

A correction to my earlier closure note: the old per-language converter implementation is obsolete after #1688, but the underlying bug is still reproducible on current `main`. A converter resolved for `fr-FR` under a German UI culture produced `dreizehn heures dreiundzwanzig`, and `en-US` produced `eins dreiundzwanzig`.

#1818 ports the minimal correct idea to the current YAML-generated phrase-clock engine, keeps the existing public converter interface compatible, and adds the explicit-culture overload plus mismatched-culture regression coverage. The replacement commit credits Julian Katzwinkel as co-author. This PR remains closed only because its implementation targets the replaced architecture and its branch cannot be updated by maintainers.

PR #1539, opened 2024-09-06
Replace "ناقص" with "سالب" for negative number representation in Arabic

omarjkf:
I updated the Arabic representation of negative numbers from "ناقص" to "سالب" to provide a more accurate translation.


omarjkf:
@dotnet-policy-service agree

PR #1580, opened 2025-06-26
feat: add English possessive inflection

### LINKED ISSUE #370: Method to make things possessive and visa-versa (Possessivize?)
```
var name = "Jesus"
var message = string.Format("It is {0} birthday!", name.Possessivize());
// message == "It is Jesus' Birthday"

var name = "David"
var message = string.Format("It is {0} birthday!", name.Possessivize());
// message == "It is David's Birthday"
```

Of course, I know different style guides treat possessive forms differently, so this would need to have a configuration option.

For instance:
Some style guides the possessive form of Illinois should be Illinois' because it ends with an s, some say it should be Illinois's because all possessive nouns get the `'s` treatment, some say it should be Illinois's because the "s" is silent, thus the "s" in the `'s' would be pronounced.

I imaging it can get even stranger if you want a possessive plural (proper) noun that ends with an s, like Jones, e.g., "Let's go to the Joneses' house".

Note: I can't think of a word denotes the opposite of being possessive.

maoalvin:
English nouns and noun phrases can now be converted to possessive form without guessing plurality from a trailing `s`. Singular nouns use `'s` by default (including `James` and `boss`), plural nouns ending in `s` use an apostrophe, and irregular plurals such as `children` use `'s`; callers can also select the apostrophe-only style for singular words ending in `s`.

The API is intentionally English-only, matching Humanizer's existing inflector surface rather than introducing a localization abstraction. Plurality and the alternate terminal-`s` style are explicit parameters because neither can be inferred reliably from the input text.

This preserves @maoalvin's original proposal and all five of their commits while modernizing the implementation for the current architecture and API conventions.

Fixes #370

## Validation

- `LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 dotnet test --project tests/Humanizer.Tests/Humanizer.Tests.csproj --minimum-expected-tests 1` — 57,346 passed on each of net8.0, net10.0, and net11.0
- `dotnet pack src/Humanizer/Humanizer.csproj -c Release -o <temp-path>`
- `dotnet format Humanizer.slnx --verify-no-changes --verbosity diagnostic`
- net48 API approval updated; net48 tests are Windows-only and were not run on WSL

## Checklist

- [x] Implementation is clean
- [x] No code was copied from StackOverflow, a blog, or another OSS project
- [x] There are very few or no comments
- [x] XML documentation is added/updated for the addition
- [x] The PR branch includes the latest commits from `main`
- [x] The issue being fixed is linked
- [x] Readme is updated for the new feature
- [x] The supported WSL test matrix, pack, and format verification pass


src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.Approve_Public_Api.DotNet10_0.verified.txt
```diff
@@ -1805,6 +1807,12 @@ namespace Humanizer
         Plural = 1,
         CouldBeEither = 2,
     }
+    public enum PossesiveSuffixOverride
```
clairernovotny, line 1810: Please use `PascalCase` for the enumeration values.

clairernovotny (dismissed):
Thank you for your contribution! 

Aside from the minor enumeration value casing, how would this work for languages other than English?

src/Humanizer.Tests/ApiApprover/PublicApiApprovalTest.Approve_Public_Api.DotNet10_0.verified.txt
clairernovotny: > Please use `PascalCase` for the enumeration values.

Not addressing in place: the suffix enum was removed entirely in 87faaa82. The current API uses explicit, conventionally named parameters instead.

src/Humanizer/Inflections/Vocabulary.cs
```diff
@@ -145,6 +152,36 @@ public void AddSingular(string rule, string replacement) =>
         return word;
     }
 
+    public string? ToPossessive(string? word, PossesiveSuffixOverride? possessiveSuffix)
+    {
+        if (word == null || string.IsNullOrWhiteSpace(word))
```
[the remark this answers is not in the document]
clairernovotny: > Null check is already done by `string.IsNullOrWhiteSpace`

Not addressing in place: the `Vocabulary.ToPossessive` method and its redundant guard were removed in 87faaa82. The current extension has a single null/whitespace guard required for nullable flow analysis across all target frameworks.

clairernovotny:
> Aside from the minor enumeration value casing, how would this work for languages other than English?

This is now explicitly English-only, matching Humanizer's existing inflector surface (`Pluralize`, `Singularize`, and related identifier transformations). Possession is not universally suffix-based, so the update deliberately avoids presenting this as a localized API or adding a locale abstraction. The enum was removed, and the public API now exposes explicit plurality and terminal-`s` style parameters.

PR #1660, opened 2025-12-30
Optimize FromUnderscoreDashSeparatedWords to reduce allocations

alexzzzs:
## Optimize FromUnderscoreDashSeparatedWords to reduce allocations

Optimizes the `FromUnderscoreDashSeparatedWords` method to eliminate unnecessary memory allocations.

The original implementation (string.Join(" ", input.Split(['_', '-']))) allocates multiple intermediate strings and arrays for each call.

### Changes

- Replaced `string.Join(" ", input.Split(['_', '-']))` with `string.Create` for .NET 7.0+ (in-place modification)
- Added `Replace` fallback for older frameworks
- Zero temporary allocations, O(N) time complexity

### Testing

- All unit tests pass
- Builds successful for .NET 8.0, 10.0, and 4.8

### DIFF
--- a/src/Humanizer/StringHumanizeExtensions.cs
+++ b/src/Humanizer/StringHumanizeExtensions.cs
@@ -32,19 +32,20 @@ public static partial class StringHumanizeExtensions
     private static Regex FreestandingSpacingCharRegex() => FreestandingSpacingCharRegexField;
 #endif
 
-   static string FromUnderscoreDashSeparatedWords(string input)
+    static string FromUnderscoreDashSeparatedWords(string input)
     {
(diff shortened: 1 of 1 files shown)


hazzik:
Why not https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.replace ?

alexzzzs:
I wrote the manual loop to handle both _ and - in a single pass to avoid iterating over the span twice.
Although, I didn't realise that MemoryExtensions.Replace uses SIMD/vectorization, which would likely make up for that.
I updated the code to use MemoryExtensions.Replace.

PR #1692, opened 2026-03-21
Fix typo in TimeOnly clock notation example

Daitbro:
Исправлена опечатка в разделе TimeOnly to Clock Notation
- Удалена лишняя обратная кавычка в примере кода
Исправлена ошибка в ByteSize
Closes #опционально_номер_issue

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -699,7 +699,7 @@ new TimeOnly(14, 30).ToClockNotation() => "duas e meia"
 
 // Round to nearest 5 minutes
 new TimeOnly(15, 7).ToClockNotation(ClockNotationRounding.NearestFiveMinutes) 
-    => "five past three"
+    => "five past three."
 ```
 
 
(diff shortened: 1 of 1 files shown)


clairernovotny:
Thanks @Daitbro for taking the time to improve the documentation. I rechecked this against current `main` and the original PR base. The proposed period would make the example disagree with the API: the [English clock template](https://github.com/Humanizr/Humanizer/blob/f8a6f583929cd82f64c2180e4d38457f22d2f135/src/Humanizer/Locales/en.yml#L454) returns `five past {hour}` without punctuation, and the [public API example](https://github.com/Humanizr/Humanizer/blob/f8a6f583929cd82f64c2180e4d38457f22d2f135/src/Humanizer/TimeOnlyToClockNotationExtensions.cs#L35) likewise documents `five past three`. The current [README example](https://github.com/Humanizr/Humanizer/blob/f8a6f583929cd82f64c2180e4d38457f22d2f135/readme.md#L701-L703) is therefore already correct. The other change, `c#` to `csharp`, is optional because both are valid Linguist aliases, as noted in the review discussion. Since neither hunk fixes a remaining issue, I’m closing this without a replacement PR. We appreciate the contribution.

PR #1789, opened 2026-05-25
fix: use CurrentCulture for default locale resolution (#1562)

### LINKED ISSUE #1562: Using incorrect culture
https://github.com/Humanizr/Humanizer/blob/d7368b9f8d916c632994128b29aed347d537b955/src/Humanizer/Configuration/LocaliserRegistry.cs#L30-L38

This code is using the `CurrentUICulture`, which is not correct.  It should be using the `CurrentCulture` instead.  While these are often set to the same value, this is not always the case -- in particular, the UI culture is often `en-US` when the culture is some other English variant.

* [`CurrentUICulture`](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentuiculture?view=net-9.0) should be used only for UI string resource translations.
* [`CurrentCulture`](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture?view=net-9.0) should be used for all other purposes, including numeric and date formatting.
* Only `CurrentCulture` reflects [user custom settings](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-globalization-cultureinfo-currentculture#the-current-culture-and-user-overrides) from the control panel.

I noticed this error when calling `ToOrdinalWords` on a `DateTime`.

wyf027:
## Summary
- Resolve null-culture lookups from `CurrentCulture` instead of `CurrentUICulture` in `LocaliserRegistry`
- Keep `ResolveForUiCulture()` explicitly bound to `CurrentUICulture`
- Update default Humanizer services (ordinalizers, date-to-ordinal converters, collection formatters, headings) to use culture-based resolution
- Align `Ordinalize` overloads and heading helpers with the same default

Fixes #1562

## Test plan
- [ ] CI passes (`LocaliserRegistryTests`, `HeadingTests`, existing localization suites)
- [ ] When `CurrentCulture` and `CurrentUICulture` differ, formatting uses `CurrentCulture`

### DIFF
--- a/src/Humanizer/Configuration/Configurator.cs
+++ b/src/Humanizer/Configuration/Configurator.cs
@@ -47,7 +47,7 @@ public static class Configurator
     public static LocaliserRegistry<ITimeOnlyToClockNotationConverter> TimeOnlyToClockNotationConverters { get; } = new TimeOnlyToClockNotationConvertersRegistry();
 #endif
 
-    internal static ICollectionFormatter CollectionFormatter => CollectionFormatters.ResolveForUiCulture();
+    internal static ICollectionFormatter CollectionFormatter => CollectionFormatters.ResolveForCulture(null);
 
     /// <summary>
     /// The formatter to be used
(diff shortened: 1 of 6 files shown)


wyf027:
Azure CI is green after correcting the en-US heading assertion (`14f3ee3f`). Ready for maintainer review.

PR #1794, opened 2026-05-25
fix: ToMetric uses rounded value for metric tier selection (#1695)

wyf027:
## Summary
- When `decimals` is specified, compute the metric exponent from the **rounded** value before choosing a prefix.
- Fixes cases like `999.9.ToMetric(decimals: 0)` returning `"1k"` instead of `"1000"`.

Fixes #1695

## Test plan
- [x] Added `ToMetric` regression case for `999.9` with `decimals: 0`
- [ ] CI green on `main`

### DIFF
--- a/src/Humanizer/MetricNumeralExtensions.cs
+++ b/src/Humanizer/MetricNumeralExtensions.cs
@@ -422,20 +422,29 @@ public static class MetricNumeralExtensions
     /// <returns>A number in a Metric representation</returns>
     static string BuildRepresentation(double input, MetricNumeralFormats? formats, int? decimals)
     {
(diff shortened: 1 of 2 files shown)


wyf027:
Pushed CI fix (`1f34a5fc`):

The previous approach rounded the raw input before exponent selection, which turned small values like `-3.91e-3` (decimals: 1) into `0` and caused `Log10(0)` / symbol index failures.

Now we only promote to the next metric tier in the no-prefix (`exponent == 0`) path when `Math.Round(input, decimals)` reaches `>= 1000` (fixes #1695 `999.9` → `1k` without breaking milli/micro cases).

wyf027:
Azure CI is green after the tier-selection fix (`1f34a5fc`). Ready for maintainer review.

Issue #349, opened 2014-11-19
CLR_EXCEPTION_System.Resources.MissingManifestResourceException in 8.1 Phone Store app


craigktreasure:
I'm getting the following crash in a Windows Phone 8.1 WinRT app that has been deployed to the store. I don't get it sideloading the appx or in the emulator.
```
Problem function: humanizer_ni!Humanizer.Localisation.Formatters.DefaultFormatter.Format	
Exception type: CLR_EXCEPTION_System.Resources.MissingManifestResourceException

"Frame    Image                                   Function                                                                                                                  Offset        
0        mscorlib.ni.dll                         System.Resources.ResourceManager.GetString                                                                                0x000002b4    
1        Humanizer.ni.DLL                        Humanizer.Localisation.Formatters.DefaultFormatter.Format                                                                 0x00000042    
2        Humanizer.ni.DLL                        Humanizer.Localisation.Formatters.DefaultFormatter.GetResourceForTimeSpan                                                 0x0000002a    
3        Humanizer.ni.DLL                        Humanizer.Localisation.Formatters.DefaultFormatter.TimeSpanHumanize                                                       0x00000010    
4        Humanizer.ni.DLL                        Humanizer.TimeSpanHumanizeExtensions+_GetTimeParts_d__4.MoveNext                                                          0x000001a8    
5        SYSTEM.CORE.NI.DLL                      System.Linq.Enumerable+WhereEnumerableIterator_1[[System.__Canon,_mscorlib]].MoveNext                                     0x0000005c    
6        SYSTEM.CORE.NI.DLL                      System.Linq.Enumerable+_TakeIterator_d__3a_1[[System.__Canon,_mscorlib]].MoveNext                                         0x00000072    
7        mscorlib.ni.dll                         System.String.Join                                                                                                        0x00000042    
8        Humanizer.ni.DLL                        Humanizer.TimeSpanHumanizeExtensions.Humanize                                                                             0x0000010e    
9        Humanizer.ni.DLL                        Humanizer.TimeSpanHumanizeExtensions.Humanize                                                                             0x00000012  
```

I'm also seeing some related warnings when i build:
```
4>MakePRI : warning 0xdef01051: No default or neutral resource given for 'Humanizer.Properties.Resources/DateHumanize_MultipleSecondsAgo_Dual'. The application may throw an exception for certain user configurations when retrieving the resources.
[...]

PR #1927, opened 2026-08-01
fix: defer exact ordinal suffix concatenation

clairernovotny:
## Summary

- defer the selected exact ordinal suffix until after the descending scale loop
- preserve first matching outer-scale, compaction, leading-one removal, output, and reference behavior
- cover the Italian suffix and outer-scale paths with a focused regression theory

## Security

Addresses [CodeQL alert #349](https://github.com/Humanizr/Humanizer/security/code-scanning/349) without adding a `StringBuilder` allocation to a loop that can append at most once.

## Validation

- focused Italian regression: 2/2 on net8.0, net10.0, and net11.0
- full Humanizer.Tests: 61,019/61,019 on each net8.0, net10.0, and net11.0
- baseline/candidate equivalence: identical 120,039-input digest and reference semantics on all three TFMs
- allocation parity: exact match for suffix, outer-scale, and mixed paths on all three TFMs
- repeated interleaved timing: no consistent material regression
- `tools/docs/generate-language-coverage.ps1 -UpdateReferenceFingerprint`: canonical no-op
- documentation `-ValidateOnly` and `-Mode Validate`
- `dotnet format Humanizer.slnx --verify-no-changes --verbosity diagnostic`
- Release pack for netstandard2.0, net48, net8.0, net10.0, and net11.0
- analyzer package consumer smoke

## Checklist

- [x] Implementation is focused and follows existing coding standards
- [x] No code-analysis or formatting warnings
- [x] Focused and full unit-test coverage passes
- [x] No copied third-party code
- [x] No public API or documentation behavior change
- [x] Based on the current live `main`

### DIFF
--- a/src/Humanizer/Localisation/NumberToWords/TriadScaleNumberToWordsConverter.cs
+++ b/src/Humanizer/Localisation/NumberToWords/TriadScaleNumberToWordsConverter.cs
@@ -187,6 +187,8 @@ class TriadScaleNumberToWordsConverter(TriadScaleNumberToWordsConverter.Profile 
     /// </summary>
     string ApplyExactScaleOrdinalTransforms(string words, long number)
     {
+        var exactOrdinalSuffix = string.Empty;
+
         // Walk from the largest scale downward so the first exact scale match wins. A smaller
         // match would be wrong here because the rewrite needs the outermost terminating scale.
         foreach (var scale in profile.Scales)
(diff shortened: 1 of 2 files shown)


clairernovotny:
Reviewed the CodeRabbit nitpick: the shared `AdditionalOrdinalCases` already contains both `it` rows (`2000` and `2000000`), and `NumberWordPhraseTests` passes the Italian culture explicitly, so `UseCulture` is not required. The focused two-row theory is intentional: it names and pins the exact suffix-vs-outermost-scale invariant for this CodeQL refactor, which the broad shared matrix does not make explicit.

